PR #4, opened 2022-10-06
Simple responsive

romulomourao:
Using max-width instead of fixed width so it won't break on smaller screens


raphamorim:
Included changes here e6b46747bb5f08308eee040efd7e1f1f9aecb155

PR #14, opened 2023-03-26
config: use XDG base directory

avelino:
Ref: https://github.com/raphamorim/rio/discussions/6#discussioncomment-5431603


raphamorim:
Perfect! Thanks @avelino

PR #65, opened 2023-05-28
gh action/release: no condition needed, 'ON' already validates the branch and tag

avelino:


raphamorim (approved):

.github/workflows/nightly.yml
```diff
@@ -35,7 +35,6 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
       - name: Release nightly
         uses: softprops/action-gh-release@v1
-        if: github.ref == 'refs/heads/main'
```
raphamorim, line 38: can we do a check to see if changes are in certain files (`*.rs`) because is overwhelming receive tags for every PR merge.

raphamorim: maybe something like this https://github.com/dorny/paths-filter ?

raphamorim: Or https://github.com/tj-actions/changed-files

raphamorim: cc @devhector (if you want to take a look as well)

avelino: That would be for another issue, not this contribution

raphamorim: Agree, approving.

raphamorim (approved):

Issue #52, opened 2023-05-25
vim background color is not correct


siimon:
Not sure exactly what the culprit is, but the colorscheme in vim is not rendered correctly if you are not using the default colorscheme, see screenshot for example

<img width="774" alt="Screenshot 2023-05-25 at 15 27 52" src="https://github.com/raphamorim/rio/assets/566235/b113c9da-6721-4ecf-a4f2-aff2b6dd73a0">

raphamorim:
Can't reproduce on my vim but my nvim is having similar issues. Probably is some missing instruction in the Rio to deal with specific configuration (my vimrc and nvim configs are different)

### My VIM

<img width="774" alt="Screenshot 2023-05-25 at 15 29 26" src="https://github.com/raphamorim/rio/assets/3630346/12cdc0c4-8be6-4abf-9f5a-e21d23970ced">

niuez:
It might cause to use true color theme. 24bit colors are rendered by following codes.

- fg
https://github.com/raphamorim/rio/blob/ec4f89dc6ae50d4ecf89183030bbbd7b1c68d544/rio/src/screen/state.rs#L122
- bg
https://github.com/raphamorim/rio/blob/ec4f89dc6ae50d4ecf89183030bbbd7b1c68d544/rio/src/screen/state.rs#L135

fg and bg are same, so we looks like no characters on screen.

I tried fixing with `ColorRGB::to_arr()`, example of function to convert to ColorArray.

```rust
impl ColorRgb {
    pub fn to_arr(&self) -> ColorArray {
        [
            self.r as f32 / 255.0,
            self.g as f32 / 255.0,
            self.b as f32 / 255.0,
            1.0
        ]
    }
}
```

Here is example.
![image](https://github.com/raphamorim/rio/assets/30189186/12900803-3291-48c2-b5e8-7ca66f54dd0d)

raphamorim:
good catch @niuez yes, rgb colors are being completely ignore (I can reproduce it as well).

Would mind send a PR correcting it in the `AnsiColor::Spec(_rgb) => self.named_colors.foreground` to use ColorRgb with to_arr as you suggested or extend ColorBuilder https://github.com/raphamorim/rio/blob/ec4f89dc6ae50d4ecf89183030bbbd7b1c68d544/colors/src/lib.rs#L371-L380

raphamorim:
@siimon can you validate if it still happens with the canary release build? https://github.com/raphamorim/rio/releases

[...]

PR #66, opened 2023-05-29
fix to render specific 24bit colors

niuez:
#52
I used `ColorBuilder` as you suggested.


raphamorim (approved):
great! thank you @niuez for finding the root cause and push the fix ⭐

Issue #67, opened 2023-05-30
The character on cursor is hidden by it.


niuez:
Hi, I want to use rio with beam cursor, but the cursor hide the character. Other kind cursors also do.
Because the cursor is rendered to replace the character to cursor character.
It should be like this.

<img width="399" alt="image" src="https://github.com/raphamorim/rio/assets/30189186/4d7cd733-2783-4e58-821b-745008fb4575">

<details>
<summary>my opinion</summary>
I think rendering the cursor by Sugarloaf's Rect is one of the solutions. 
The image above was created by this code. Cursors are considered to be associated with characters, so I defined new SugarStyle `on_cursor`.

https://github.com/niuez/rio/blob/0ea71b80d30af2038fa0a361b4d73b4d92be6b24/sugarloaf/src/sugarloaf.rs#L299-L307

But there are some problems.

- Sugarloaf is rendering engine. It should not be contain cursor shape information. The cursor shapes should be defined in rio.
- The color of cursor should be able to change by rio.
- How to render more complex cursors. For example, unfilled rectangle.
</details>

raphamorim:
Yea, that's for bring this up. Unfortunately the way sugarloaf is built now, consider cursor as a sugar so it would replace a text block, it would work for Block cursors but not for other cases. For fix this issue I was thinking in few possibilities:

- add a new type called `SugarPileLayer` that would be used by `SugarPile` to contain multiple layers (that would be useful for rects that are rendered in background, so after the refactor will be 3 layers: rect, text, text[cursor or special]).
- add a new type called `SugarLayer` that would be used by `Sugar`. Kinda same approach as `SugarPileLayer` but for sugar blocks only.
- render it like we do with `pile` method (in the sugarloaf examples), would be an external method that we say we want a sugar in a specific position in the screen.

raphamorim:
I kinda of like SugarLayer cause it simplify stuff (the way we consider Rects for example).

Maybe something like this:

```rust
pub struct SugarLayer {
    pub layers: Vec<char>
}

#[derive(Debug)]
pub struct Sugar {
    pub content: SugarLayer,
    pub foreground_color: [f32; 4],
    pub background_color: [f32; 4],
    pub style: Option<SugarStyle>,
}
```

niuez:
Using `SugarLayer` becomes block and underline cursor OK, but Beam will be like this I think.
[...]

PR #74, opened 2023-05-31
Example of rendering cursors by SugarDecoration

niuez:
#67
Sorry it contains feature for changing cursor by ANSI escape sequences for confilm...
We opened the pull request diff as we thought it would be easier to explain the feature to render cursor.

The feature required for rendering cursor in Sugarloaf is SugarDecoration and the code to draw it.
`position` and `size` of Decoration represent percentage of width and height of glyphs. This is rendered at the same time as the corresponding character in `stack()`


raphamorim:
When the sugar has bigger width this is what happens:

Screenshot 2023-05-31 at 18 32 59
Screenshot 2023-05-31 at 18 33 11

But looking other gpu terminals like Alacritty, looks they also do same way.

Screenshot 2023-05-31 at 18 33 20
Screenshot 2023-05-31 at 18 33 26

raphamorim (approved):

PR #77, opened 2023-06-02
fix: support to clipboard in linux

joseemds:
I built rio in linux (with xorg), but when i tried to use the clipboard i was receiving some errors from console:

```
$ cargo build --release
$ ./target/release/rio
Attempting to get the contents of the clipboard, which hasn't yet been implemented on this platform.
```
While investigating the problem i found that the `copypasta` dependency was with `default-feature = false`, i tried both removing that option and adding explicit `features = ["x11", "wayland"] but i was receiving the same error.

What i did to fix was to add the `features` field in `rio/Cargo.toml` and built it with `cargo build --release --features=x11` and now seems to work perfectly.

quick note: Was only tested in xorg.

Please let me know if any change is needed or if this is not the best fix!

obs: i runned make lint and returned some errors even but i didnt edited any file with code, so i left it as it was.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,6 +383,8 @@ dependencies = [
  "objc",
  "objc-foundation",
  "objc_id",
+ "smithay-clipboard",
+ "x11-clipboard",
 ]
 
 [[package]]
@@ -738,6 +740,16 @@ name = "futures-sink"
 version = "0.3.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f43be4fe21a13b9781a69afa4985b0f6ee0e1afab2c6f454a8cf30e2b2237b6e"
+
+[[package]]
+name = "gethostname"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c1ebd34e35c46e00bb73e81363248d627782724609fe1b6396f553f68fe3862e"
+dependencies = [
+ "libc",
+ "winapi 0.3.9",
+]
 
 [[package]]
 name = "getrandom"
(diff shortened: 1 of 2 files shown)


raphamorim:
Hey @joseemds thank you for the pull request, yes it does make sense. I looked the alacritty source code and looks that they also use same fix https://github.com/alacritty/alacritty/blob/4b92388396c6da783809c7ce511b5cf6be8a0624/alacritty/Cargo.toml#L79-L87

The only thing is that we need to update clipboard.rs to look similar to their solution https://github.com/alacritty/alacritty/blob/4b92388396c6da783809c7ce511b5cf6be8a0624/alacritty/src/clipboard.rs#L1-L16

e.g:

```rust
#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
use std::ffi::c_void;

use log::{debug, warn};

use alacritty_terminal::term::ClipboardType;

#[cfg(any(test, not(any(feature = "x11", target_os = "macos", windows))))]
use copypasta::nop_clipboard::NopClipboardContext;
#[cfg(all(feature = "wayland", not(any(target_os = "macos", windows))))]
use copypasta::wayland_clipboard;
#[cfg(all(feature = "x11", not(any(target_os = "macos", windows))))]
use copypasta::x11_clipboard::{Primary as X11SelectionClipboard, X11ClipboardContext};
#[cfg(any(feature = "x11", target_os = "macos", windows))]
use copypasta::ClipboardContext;
use copypasta::ClipboardProvider;
```

Otherwise it will fail for other cases like is failing in the Ubuntu CI now. Would you mind include it in this PR?

joseemds:
Hi @raphamorim sorry for the delayed response, i compared `rio/clipboard.rs` with `alacritty/clipboard.rs`  and they look similar, with just some namespace and formatting differences.

Indeed running make lint locally fails as in the CI. I noticed that the command for clippy in the `Makefile` is  `cargo clippy --all-targets --all-features -- -D warnings` and i guess thats the root of the problem.

Running `cargo clippy` without `--all-features` or with `--features=x11` satisfies clippy.

Maybe there is something in clipboard that i couldn't catch, if so please let me know!

But i think that the solution would be to have different targets for windos, macos and linux in Makefile? 

What do you think?

raphamorim:
Oh yes, you are very right about the clippy. I think remained as a legacy form the beginning, please remove the `--all-features` and leave the `--all-clients` there. Should be enough as you said. 👌

raphamorim:
Hey @joseemds, I took your changes and included in the main branch with the suggested fixes https://github.com/raphamorim/rio/commit/45c97186d1057f6201c8732b9273bf92f39d95d5. Thank you for the contribution 🎖️ ⭐

PR #155, opened 2023-07-20
add application id to window on x11/wayland

tranzystorek-io:
Not sure how many apps integrate with this change, but at least tested that it enables Waybar on wayland to detect rio's desktop file and icon.

### DIFF
--- a/rio/src/screen/constants.rs
+++ b/rio/src/screen/constants.rs
@@ -24,3 +24,9 @@ pub const DEADZONE_END_Y: f64 = -2.0;
 
 #[cfg(target_os = "macos")]
 pub const DEADZONE_START_X: f64 = 80.;
+
+#[cfg(all(
+    any(feature = "wayland", feature = "x11"),
+    not(any(target_os = "macos", windows))
+))]
+pub const APPLICATION_ID: &str = "rio";
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix unconditional query of xdg-portal settings on Wayland
 - Fix `Maximized` startup mode not filling the screen properly on GNOME Wayland.
 - Fix Default Vi key bindings for `Last`/`First` actions not working on X11/Wayland.
+- Set `app_id`/`WM_CLASS` property on Wayland/X11
 
 ## 0.0.12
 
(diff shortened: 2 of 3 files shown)


rio/src/screen/constants.rs
```diff
@@ -24,3 +24,9 @@ pub const DEADZONE_END_Y: f64 = -2.0;
 
 #[cfg(target_os = "macos")]
 pub const DEADZONE_START_X: f64 = 80.;
+
+#[cfg(all(
+    any(feature = "wayland", feature = "x11"),
+    not(any(target_os = "macos", windows))
+))]
+pub const APPLICATION_ID: &str = "rio";
```
raphamorim, line 32: Does Rio with uppercase `R` works as well?

raphamorim: Asking because first letter uppercased is default for other platforms.

and alacritty also does have uppercased for wayland/x11 https://github.com/alacritty/alacritty/blob/master/alacritty/src/config/window.rs#L19C11-L19C23

tranzystorek-io: Waybar looks for desktop files based on the app_id (https://github.com/Alexays/Waybar/blob/3ecd4030e3318863d2bab087d01c9921624f0269/src/modules/wlr/taskbar.cpp#L66C1-L90C1), so we'd probably have to rename the desktop file to match.

raphamorim: Asking because Rio with uppercased `R` is the default for other platform.

and alacritty also sets with first letter uppercased for x11/wayland https://github.com/alacritty/alacritty/blob/master/alacritty/src/config/window.rs#L19

raphamorim: https://github.com/alacritty/alacritty/blob/2101d5ac37d65114781a84b889d96eafe768860a/extra/linux/Alacritty.desktop#L4

tranzystorek-io: > and alacritty also sets with first letter uppercased for x11/wayland https://github.com/alacritty/alacritty/blob/master/alacritty/src/config/window.rs#L19

Alacritty might be an outlier:

obraz

tranzystorek-io: > Does Rio with uppercase `R` works as well?

Yeah, Waybar seems to pick it up no matter the casing.

raphamorim: hmmm, let's keep lowercased then to match with wezterm and kitty

raphamorim: Did you check if for multiwindow it works without any inconsistency?

tranzystorek-io: > Did you check if for multiwindow it works without any inconsistency?

Yup, all windows have the app_id, and all get an icon in Waybar.

raphamorim (approved):

PR #166, opened 2023-08-01
Set font family of code blocks

grantholle:
Right now it's inheriting from the reset, `Helvetica`, which makes a poor coding font 😄


raphamorim (approved):

raphamorim:
Thanks @grantholle

PR #174, opened 2023-08-03
change .deb package name to avoid conflicts

### LINKED ISSUE #173: Debian package rio overwrites manually installed rio terminal .deb pkg
There is a package called "rio" available in the Debian repositories, accessible through this link: [rio debian repo](https://packages.debian.org/bookworm/rio).

However, there is an issue when updating the system using 'sudo apt update'. When doing so, Debian will install the rio package from the repositories, but unfortunately, it removes the Rio terminal.
I think there's some workarounds, but this is the default behavior right now.

DanielTolentino:
I don't know if I'm changing the right place, but looks like a good start!

Fix #173

### DIFF
--- a/rio/Cargo.toml
+++ b/rio/Cargo.toml
@@ -68,14 +68,14 @@ wayland = [
 ]
 
 [package.metadata.deb]
-name = "rio"
+name = "rio-term"
 maintainer = "Raphael Amorim <rapha850@gmail.com>"
 copyright = "2023, Raphael Amorim <rapha850@gmail.com>"
 depends = "$auto"
 section = "admin"
 priority = "optional"
 assets = [
-    ["../target/release/rio", "usr/local/bin/", "755"],
+    ["../target/release/rio-term", "usr/local/bin/", "755"],
     ["../misc/rio.desktop", "usr/share/applications/", "644"],
     ["../misc/rio.terminfo", "usr/share/info/", "644"],
     ["../misc/logo.svg", "usr/share/icons/hicolor/scalable/apps/rio.svg", "644"],


rio/Cargo.toml
```diff
@@ -68,14 +68,14 @@ wayland = [
...
+name = "rio-term"
 maintainer = "Raphael Amorim <rapha850@gmail.com>"
 copyright = "2023, Raphael Amorim <rapha850@gmail.com>"
 depends = "$auto"
 section = "admin"
 priority = "optional"
 assets = [
-    ["../target/release/rio", "usr/local/bin/", "755"],
+    ["../target/release/rio-term", "usr/local/bin/", "755"],
     ["../misc/rio.desktop", "usr/share/applications/", "644"],
```
raphamorim, line 79: Have you test it in the debian?
wonder if needs to be `rio-term.desktop`

rio/Cargo.toml
```diff
@@ -68,14 +68,14 @@ wayland = [
...
-name = "rio"
+name = "rio-term"
 maintainer = "Raphael Amorim <rapha850@gmail.com>"
 copyright = "2023, Raphael Amorim <rapha850@gmail.com>"
 depends = "$auto"
 section = "admin"
 priority = "optional"
 assets = [
-    ["../target/release/rio", "usr/local/bin/", "755"],
+    ["../target/release/rio-term", "usr/local/bin/", "755"],
```
otavio, line 78: Why `usr/local/bin`?

otavio (changes requested):
What conflicts with rio?

rio/Cargo.toml
DanielTolentino: I'm getting this error, and I can't solve it:
`cargo-deb: Asset file path does not match any files: /home/daniel/git-src/dev/rio/rio/../target/release/rio-term`
I've change the assests names, changing only the "name" property doens't work when installing the package, the system still install as "rio".

DanielTolentino:
> What conflicts with rio?

See #174

rio/Cargo.toml
DanielTolentino: It's already the default path, I didn't change it.

rio/Cargo.toml
raphamorim: `target/release/rio-term` and `rio.desktop` does work then?

DanielTolentino: No, `target/release/rio-term` and` rio.desktop `or `rio-term.desktop` doesn't build at all, getting that path error thing.
It only builds if only change
``` toml
[package.metadata.deb]
name = "rio-term"

```

But this doesn't fix the apt naming conflict, still build as rio

raphamorim:
Hey @DanielTolentino any reason to close it? I was planning to review after the vacation

DanielTolentino:
All the changes I did both in this PR's commits and locally didn't work. The package kept being built as "rio", I can't figure out how to properly make the changes and I'm not interested in trying anymore.

raphamorim:
Gotcha! will take a look on the issue once I come back

Issue #183, opened 2023-08-15
Linux Default Keybind clashes with readline


cwegener:
The default keybinds clash with readline, preventing the readline functions from working when running a program inside rio that uses readline (e.g. emacs, or any shell)

Cheatsheet of important readline keyboard shortcuts: https://www.bigsmoke.us/readline/shortcuts

The specific keybinds that clash are:

- `Control + N`
- `Control + W`
- `Control + T`

This wouldn't be a big problem if these could be changed via configuration.

Should the default keybinds be changed to more sensible ones instead?

**EDIT**: Forgot to include the configuration settings that I tried out to disable the conflicting keybinds, but that configuration did not result in disabling the default keybinds:

```
[bindings]
keys = [
	{ key = "n", with = "control", action = "none" },
	{ key = "t", with = "control", action = "none" },
	{ key = "w", with = "control", action = "none" }
]
```

raphamorim:
Hey @cwegener thank you for the issue, very good point with readline keyboard. Considering is Control + key I suppose you are using linux right?

I wonder if the best solution would be change the defaults for `control + shift`, like `n + cntrl + shift`.

> EDIT: Forgot to include the configuration settings that I tried out to disable the conflicting keybinds, but that configuration did not result in disabling the default keybinds:

Could you please try again with latest main, believe fixed the overwrite in the latest changes.

cwegener:
Changing the modifier to `control + shift` sounds like a good approach. There could be some other possible clashes when using `control + shift`  (e.g. the default keybinds for popular Desktop Environments may possibly be using `control + shift` for some of their global functions), but I think those clashes are a lower priority than the clashes with readline.

Yes, this is for Linux.

Btw. The config snippet I was testing with actually wasn't getting parsed successfully by `rio` for some reason (not sure why). So, that might have been why I wasn't able to override the defaults with my configuration.

I'll try with the latest main and report back.

Oh. And the rendering is really fast. (It's my first time testing `rio` and I was pretty impressed with the rendering speed. It definitely feels faster than any other option I've tried on Wayland ... and I think I've tried them all ... Currently using `wayst` as my main terminal because it's the least sucky one of them all ... )

cwegener:
Ah. `action="ReceiveChar"` is what I need. All good now.

Changing the default modifier on Linux to `control` + `shift` would still be a good idea, considering how commonly readline is used (it's pretty much used by every shell and every REPL that I can think of)

[...]

PR #187, opened 2023-08-18
feat!: enable readline keybinds

cwegener:
Some of the original rio default keybinds for Linux & Windows were
conflicting with the default keybinds of the readline library, thereby
prohibiting the use of those readline keybinds in programs that make use
of readline (most shells, most REPLs, most editors)

This change updates the default modifiers in rio from just `CONTROL`
to `CONTROL` + `SHIFT` for those keybinds that would otherwise conflict
with the readline ones.

Also, the documentation for default Copy & Paste keybinds on Windows and
Linux has been corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Support lowercased action and fix overwrite for actions in custom key bindings.
 - Added `ReceiveChar` action for custom key bindings.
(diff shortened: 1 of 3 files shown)


raphamorim (approved):
Great! Thank you @cwegener 
Closing #183

PR #210, opened 2023-09-09
fix: change package name from rio to rioterm

marrcelo:
Updating the package name from rio to rioterm.


Makefile
marrcelo, line 1: I was not sure if I should update the `bump-brew` to:
```
bump-brew:
	brew bump-cask-pr rioterm --version ${version}
```
So I did not update it.

raphamorim (approved):
Thank you

Issue #225, opened 2023-09-14
Doesn't work with touchscreen


androw:
I'm using rio on openSUSE linux.

I'm unable to select text on the terminal by using the touchscreen.
This is working fine on alacritty.

This is a low priority issue I guess.

raphamorim:
Oh yes, Rio doesn't have support for touch events but I would be super keen to add. I don't have any touchscreen device at the moment so will take a moment to look it.

If you want (or anyone) that would be keen to create a PR adding support, can follow up Alacritty touch strategy https://github.com/alacritty/alacritty/blob/34b5dbacd28cd1abaedf1d81cc0ebe57aa44a086/alacritty/src/input.rs#L801-L911

androw:
I'll try to look into it

raphamorim:
Also connected to the link I've sent https://github.com/alacritty/alacritty/blob/master/alacritty/src/event.rs#L1297

raphamorim:
Closing it after #226 implementation

Issue #182, opened 2023-08-15
click on cursor to highlight seems to be one or two characters off


Jinkxed:
When clicking to select and highlight on Mac OSX.  The actual highlighting starts one or two digits to the right of the cursor.

Brianalmeida:
I am seeing this too on Linux too

raphamorim:
woof, I can reproduce as well. marking as bug

Brianalmeida:
Using IBM Plex Mono, size 28

raphamorim:
And this is also applied whatever start to select text right?

Jinkxed:
Yup

Brianalmeida:
@raphamorim I noticed that the cursor highlighting is a bit better in the latest release

raphamorim:
@Brianalmeida @Jinkxed can you guys test in latest main?

Brianalmeida:
Compiling and testing now

Jinkxed:
Looking now.

raphamorim:
@Jinkxed @Brianalmeida could you try with latest main? (now is using fixed width for each cell)

Jinkxed:
Installed will play with it today and let ya know!

Brianalmeida:
@raphamorim just got back, so I'll give this another shot tomorrow morning and report back any findings

Brianalmeida:
@raphamorim using commitmono, the highlight looks good. IBM Plex Mono, and JetBrains Mono look good, too.

fdncred:
I just compiled the latest main on Windows and it seems worse now. It's 2 off instead of 1 off.

In the screenshot below, you can see my mouse selection cursor between the N and G in CHANGELOG.md. I started selecting a little bit to the right of the md and it didn't select the md and it seems to be 2 characters ahead to the left, as I drag from right to left.
![image](https://github.com/raphamorim/rio/assets/343840/e0a5ebe9-6742-4275-8427-737cd8621c17)

raphamorim:
hmmm, interesting @sztomi . I suppose you use `padding-x` equal to zero as well right?

raphamorim:
Basically rewrote the mouse position function for retrieve column on mouse position. Below you can see some examples, it's still not fully perfect but matches much better than was before

## Fira code, scale 1, padding 10, font-size 32px

[...]

PR #226, opened 2023-09-16
WIP: add basic touch support

androw:
#225 
First work on it
Need winit upgrade because of https://github.com/rust-windowing/winit/issues/3035

@raphamorim can you point me to the correct way to send click and cursor move event please?

### DIFF
--- a/frontends/cross-winit/src/screen/mod.rs
+++ b/frontends/cross-winit/src/screen/mod.rs
@@ -13,6 +13,7 @@ mod messenger;
 mod mouse;
 mod navigation;
 mod state;
+pub mod touch;
 
 use crate::clipboard::{Clipboard, ClipboardType};
 use crate::crosswords::{
@@ -33,6 +34,7 @@ use crate::screen::{
     bindings::{Action as Act, BindingKey, BindingMode, FontSizeAction},
     context::ContextManager,
     mouse::{calculate_mouse_position, Mouse},
+    touch::TouchPurpose
 };
 use crate::selection::{Selection, SelectionType};
 use core::fmt::Debug;
@@ -100,6 +102,7 @@ pub struct Screen {
     clipboard: Clipboard,
     pub modifiers: Modifiers,
     pub mouse: Mouse,
+    pub touchpurpose: TouchPurpose,
     pub ime: Ime,
     pub state: State,
     pub sugarloaf: Sugarloaf,
@@ -229,6 +232,7 @@ impl Screen {
             ime,
             sugarloaf,
             mouse: Mouse::new(config.scroll.multiplier, config.scroll.divider),
+            touchpurpose: TouchPurpose::default(),
             state,
             bindings,
             clipboard,
@@ -269,6 +273,10 @@ impl Screen {
                 self.sugarloaf.layout.scaled_sugarheight,
             ),
         )
+    }
+
+    pub fn touch_purpose(&mut self) -> &mut TouchPurpose {
+        &mut self.touchpurpose
     }
 
     #[inline]
--- a/frontends/cross-winit/src/screen/touch.rs
+++ b/frontends/cross-winit/src/screen/touch.rs
@@ -1,0 +1,265 @@
+// Part of this file was originally taken from Alacritty
+// https://github.com/alacritty/alacritty/blob/34b5dbacd28cd1abaedf1d81cc0ebe57aa44a086/alacritty/src/input.rs
+// which is licensed under Apache 2.0 license.
+
+use std::time::Instant;
+use std::{collections::HashSet, mem};
+use std::collections::hash_map::RandomState;
+
+use winit::event::{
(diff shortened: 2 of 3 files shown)


raphamorim:
Hey @androw,

That's great! So for the mouse Rio uses `mouse_position` function that does return a `Pos` https://github.com/raphamorim/rio/blob/65ffe901fdb3353e54860bc0daec8fdc4c286aff/rio/src/screen/mod.rs#L227-L250

With this `Pos` it uses `mouse_report` function that reports https://github.com/raphamorim/rio/blob/65ffe901fdb3353e54860bc0daec8fdc4c286aff/rio/src/screen/mod.rs#L1399-L1433

But I think for your case, you will need to use `on_left_click` function https://github.com/raphamorim/rio/blob/main/rio/src/screen/mod.rs#L1247-L1276 it receives the `Pos`, I think the biggest challenge here is translate touch into `Pos` (that's why I sent the mouse_position function before)

Note the mouse has a state called ClickState, it can be a click, double or triple. The click state is defined here https://github.com/raphamorim/rio/blob/main/rio/src/sequencer.rs#L438-L474

I think would be worth to make touch reuse the clickState though, then doesn't need to duplicate logic.

Keep in minde that `mouse_position` function is still not the best and needs to be fixed since whenever you change fonts it can lead to some issues (ref #182)

androw:
If someone want to test and add feedback. We should have some basics working on X11 (for Wayland we need to wait for winit next version):
- select text
- zoom in/out
- scroll

Nyabinary:
What version of Winit is needed, more specifically what commit? Because Rio uses the latest beta ver of Winit.

androw:
> What version of Winit is needed, more specifically what commit? Because Rio uses the latest beta ver of Winit.

We need this commit https://github.com/rust-windowing/winit/commit/c2ed160e639ccaa10c6f0219e2cb25901aa559be which is not included in latest beta build.

raphamorim:
@androw since we are already using the beta, I wouldn’t mind add winit dependency pointing to a git hash for the next build. Since in my experience Winit releases can take a while to wrap up (I know they are also considering a major refactor of how the event loop will work)

Nyabinary:
That's seems like a solid idea tbh, could take advantage of some new features winit too if you point to the latest git hashes. Of course, it'll need proper testing, which I'll be happy to help with :)

raphamorim:
@androw just a fyi, updated to winit 0.29.2 in main branch

androw:
Working on Wayland without patch now!
But simple text selection is not working anymore. I might check this over the weekend

raphamorim:
Niceee! thanks @androw let me know anything can help with :D and great job with it

Nyabinary:
What the status on this?

androw:
I've updated the branch and selection is still not working. I didn't have time to work on it and won't be able for some times.

raphamorim:
No stress @androw, take your time on it 🙏 
@Nyabinary please don't comment like this, feels as pressure to release. It will come naturally 🙏

Nyabinary:
> No stress @androw, take your time on it 🙏 @Nyabinary please don't comment like this, feels as pressure to release. It will come naturally 🙏

Oh, I was under the impression that it was ready to review, sorry! Should mark it as a draft PR too.

raphamorim:
No stress @Nyabinary 🙏

androw:
@raphamorim i do not fins what is wrong with my way of emulating a click. Can you have a look when you have some time?

raphamorim:
> @raphamorim i do not find what's wrong with my way of emulating a click. Can you have a look when you have some time?

Of course, you are talking about the CI error?

androw:
No (I haven't check the CI error yet).
For example, I'm using the following code to simulate a click
```
                route.window.screen.mouse.x = x;
                route.window.screen.mouse.y = y;

                let now = Instant::now();
                route.window.screen.mouse.last_click_timestamp =
                                            now;
                route.window.screen.mouse.last_click_button = MouseButton::Left;

                route.window.screen.mouse.click_state = ClickState::Click;
                route.window.screen.mouse.left_button_state = ElementState::Pressed;
                route.window.screen.on_left_click(route.window.screen.mouse_position(0));
```
and the following for releasing the click
```
            route.window.screen.mouse.click_state = ClickState::None;
            route.window.screen.mouse.left_button_state = ElementState::Released;
```
It used to work but it isn't (for example selection doesn't) since some changes on the main branch.
I think I'm not doing this the right way...

raphamorim:
> It used to work but it isn't (for example selection doesn't) since some changes on the main branch.

Got it! Will take a look once I wrap the sixel protocol 🙏

raphamorim:
@androw I've merged your changes into main (https://github.com/raphamorim/rio/commit/ae776054903e44de7b58465322eb6773d31e50ea). Overall the code does look fine for me (just added few updates in another commit), haven't tested properly the selection only the touch events but I will test before include to next release. Thank you for your work with it 🙏

PR #227, opened 2023-09-17
feat: option for ignoring theme foreground selection color

hougesen:
Using the same color when selecting as the existing foreground color can be desired, since the color can include information about the text. 

This PR aims to add a configuration option for easily turning that feature on. By default the option is disabled. 

With the option set to false: 

image

With the option set to true: 

image

I apologize if this is not a desired feature, or if it already exists. I looked around the project but was unable to find anything about this.

### DIFF
--- a/rio-config/src/defaults.rs
+++ b/rio-config/src/defaults.rs
@@ -95,6 +95,12 @@ cursor = '▇'
 #
 blinking_cursor = true
 
+# Ignore theme selection foreground color
+#
+# Default is false
+#
+ignore_theme_selection_fg_color = false
+
 # Performance
 #
 # Set WGPU rendering performance
--- a/rio/src/screen/state.rs
+++ b/rio/src/screen/state.rs
@@ -9,7 +9,7 @@ use crate::screen::{context, EventProxy};
 use crate::selection::SelectionRange;
 use rio_config::colors::{
     term::{List, TermColors},
-    AnsiColor, Colors, NamedColor,
+    AnsiColor, ColorArray, Colors, NamedColor,
 };
 use rio_config::Config;
 use std::collections::HashMap;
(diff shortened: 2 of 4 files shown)


raphamorim:
I liked the idea!

Very interesting to leave as an option since terminals implement it in different ways

Screenshot 2023-09-17 at 10 24 29
Screenshot 2023-09-17 at 10 24 19

@hougesen can you update the docs file regarding this change https://github.com/raphamorim/rio/blob/main/docs/docs/configuration-file.md

rio/src/screen/state.rs
```diff
@@ -361,6 +317,60 @@ impl State {
         stack
     }
 
+    fn fg_square_to_color_array(&self, square: &Square) -> ColorArray {
```
raphamorim, line 320: can you add an inline to this function
```rust
#[inline]
fn fg_square_to_color_array(&self, square: &Square) -> ColorArray {
```

rio-config/src/defaults.rs
```diff
@@ -95,6 +95,12 @@ cursor = '▇'
 #
 blinking_cursor = true
 
+# Ignore theme selection foreground color
+#
+# Default is false
+#
+ignore_theme_selection_fg_color = false
```
raphamorim, line 102: I think the default is true no? https://github.com/raphamorim/rio/pull/227/files#diff-b64bebdc21e5973e8c9ff4e193b6533793ef05ec19c6c812fbcf02be5367dafaR360

hougesen: Yes, that is a mistake from my side. 

The default should of course be false, so it doesn't break existing behavior. 

I have just pushed a fix.

raphamorim: perfect! 🚀

raphamorim (approved):
Thank you! Just a FYI might be I will rename this property name before release 0.0.19 since I am working in a new window API

hougesen:
Yes, no worries. 

It might also make sense to move it the a subsection of the configuration file like window, or directly in the theme. 

I was unsure where to place the configuration option so I just went with what felt easiest 😅

PR #235, opened 2023-09-19
refactor: enable CPU-specific optimizations on aarch64-apple-darwin

lovesegfault:
See the comment for context, this should enable the compiler to produce better
code for ARM macs


raphamorim (approved):
Thank you 👍

Issue #229, opened 2023-09-18
Add predefined `ToggleFullscreen` Action


hustcer:
Thanks for the great work,
Can you please add a predefined `ToggleFullscreen` Action? So I can bind custom key bindings to it

raphamorim:
Of course! marking it as a wishlist issue

sonbui00:
Is this `Command + Control + F` on MacOS?

raphamorim:
Command + Control + F (on Big Sur and earlier) or Fn + F if your Mac is running macOS Monterey or later

raphamorim:
Let’s stick with Fn + F? (Since is the new default shortcut)

hustcer:
Both `Command + Control + F` and ` Fn + F` don't work for me.
macOS Ventura  Version 13.5.2

PR #244, opened 2023-09-21
fix: Command + H can't hide rio on macOS #240

### LINKED ISSUE #240: `Command + H` can't hide rio on macOS
`Command + H` can't hide rio on macOS which is the expected behavior

$> rio --version
rioterm 0.0.19

sonbui00:
Fix #240 
Command + H can't hide rio on macOS


raphamorim:
Looks good @sonbui00 
Question: would you like to take the opportunity and implement the toggle full screen in another PR (#229)

sonbui00:
@raphamorim Yes. let me check

rio/src/sequencer.rs
```diff
@@ -331,6 +332,10 @@ impl Sequencer {
                                     route.window.winit_window.select_previous_tab();
                                 }
                             }
+                            #[cfg(target_os = "macos")]
```
raphamorim, line 335: I think hide is universal (besides wayland) no?

raphamorim: Technically would not need the target_os

sonbui00: key binding for this is only for MacOS and underline function (hide_application()) also implement for Macos only
If remove target_os filter, I think it still can run normally due to no key bind to call this

raphamorim: Ah I see, I realised now that you are using hide_application, I thought hide also would work with https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_visible set_visible (that works for almost all platforms)

raphamorim: Maybe hide can be global and if isn’t macos then use set_visible instead

sonbui00: @raphamorim just try set_visible in MacOS, it's different with hide function that expect:
- hide: can back by alt + tab or icon on dock
- set_visible: don't how to back, maybe need call set_visible(true) to take it back

If you want I can add set_visible for other platform, so someone can bind on this

raphamorim: Hmmm i see, well let’s implement only macos then and later we can revist it

raphamorim (approved):

raphamorim:
Thank you @sonbui00

PR #275, opened 2023-10-06
updating rust-version to 1.73.0

igorvieira:
## Evidence

image


tranzystorek-io:
I'm not sure we want to eagerly bump rust-version on each rustlang release, this field serves as a MSRV indicator which should be relatively low to allow less frequently updated envs

raphamorim:
Although I agree with @tranzystorek-io , I want to use this https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html#thread-local-initialization lol (as some projects as well are also updating to 1.73).

Also good to highlight that Rio doesn't really follow a MSRV (and maybe I should add this in the docs), while I do think MSRV is super useful for libraries, I feel that for applications it comes with the tradeoff of cost resources/tools/perf/compilation for users (considering that most of the users use the binary only). Maybe a good idea is to keep 1-2 versions from latest though.

raphamorim (approved):

otavio:
It is my belief that we should refrain from upgrading the Rust compiler unless it is absolutely necessary, as it can cause issues with upgrading Rio in various distributions. For instance, in NixOS, we are currently working on incorporating Rust 1.73.0 into our system. This upgrade requires us to update the underlying LLVM, and some packages are still in the process of being fixed to accommodate it.

While the new thread-local feature that doesn't use the lambda functions is a nice addition, it would have been better to postpone it until a later time, in order to make it easier for distributions to upgrade.

PR #277, opened 2023-10-09
Fix link to build from source in Linux install

krishna-kashyap:


docs/docs/install/linux.md
```diff
@@ -10,4 +10,4 @@ Installation options:
 - Nix package: [NixOS/nixpkgs/rio](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/terminal-emulators/rio/default.nix)
 - openSUSE package: [openSUSE:Factory/rioterm](https://software.opensuse.org/package/rioterm)
 
-In case your distro doesn't have the package manager option listed above, proceed to [build from the source](#build-from-the-source).
+In case your distro doesn't have the package manager option listed above, proceed to [build from the source](build-from-source).
```
raphamorim, line 13: I think should be `/docs/install/build-from-source/` no?

krishna-kashyap: I tested by modifying it from dev tools and this seemed to work, possibly because the URL for this page is already `/docs/install/linux`? can update it to do the full path though

raphamorim: yes, editing with dev tools it does work but I believe that just `build-from-source` will not gonna work later for the static page builder using [Docusaurus](https://docusaurus.io/). Can you update to the full path?

krishna-kashyap: updated

raphamorim (approved):
Thank you @krishna-kashyap

PR #285, opened 2023-10-14
Navigation: Remove magic strings

TornaxO7:
That should make it bit more robust, since you will only need to change the strings in the constants hence not needing to retype the string all type time.


raphamorim:
I like the approach, just need to fix the tests 👌

raphamorim (approved):

raphamorim:
⭐

Issue #278, opened 2023-10-09
Scroll to fast in vim and nano


DanielTolentino:
Hi, I've encountered this issue on multiple Linux distributions while using Wayland with both a touchpad and a mouse. Is there any configuration change I can make to make it behave like other terminals? Thanks.

raphamorim:
Hey @DanielTolentino thanks for the issue,

Just to confirm, you meant scroll too fast right? If yes, then I think maybe writing a configuration property to modify scroll speed.

DanielTolentino:
The scrolling in rio is too fast, what config can I change to make it properly behave like other terminals?
I can't record the screen right now

raphamorim:
Ah gotcha! ok, there's no configuration yet to control it.

Although I think is a simple change, might need to tweak this code https://github.com/raphamorim/rio/blob/main/rio/src/screen/mouse.rs#L36 to make multiplier comes from configuration file.

DanielTolentino:
Nice. I don't know if I can make a working PR on this, but I'll try to change the multiplier and build from source afterwards.
Thanks!

raphamorim:
If you would like to create a PR for it, feel free to create a draft version and we can iterate together :D

DanielTolentino:
Awesome, I'll work on it soon after I get home 😊

inudev5:
@raphamorim @DanielTolentino what should the default modifier value be? any suggestions?

inudev5:
the scroll multiplier is being applied in the shell, i can confirm. 
but the scroll behavior is not applied inside any programs or editors. i think it should be applied inside other programs also. 
@raphamorim

raphamorim:
Part of the problem of the quick scroll as well is that rio does render as soon is possible, probably we will need to move scroll to use scheduler.rs file (as have for render)

raphamorim:
Added a new property to next version:

```toml
[scroll]
multiplier = 1.0
divider = 2.0
```

scroll calculation will be based on `(accumulated scroll * multiplier / divider)` so if you want quicker scroll, keep increasing the multiplier if you want to reduce you increase the divider. Can use both properties also to find the best scroll for you

Example below with multiplier 1.0 and divider 2.0

![scroll-mul-1-div-2](https://github.com/raphamorim/rio/assets/3630346/01ef7f01-5eaa-4016-b79f-7a0ab0c1550e)

Note: This should fix for canonical terminal behavior, for tmux/nvim/vim cases I am still working around

[...]

Issue #267, opened 2023-10-02
[Keybinding] ScrollLineDown/ScrollLineUp actions not work as expected


lost22git:
os: win11
version: v0.0.22

![cc](https://github.com/raphamorim/rio/assets/65008815/2675c992-46a1-4a0c-afee-9e3c00c23d5d)

## problem
* ScrollLineUp is only triggered once
* ScrollLineDown behave same as ScrollToBottom

## config
```toml
[bindings]
keys = [
  { key = "=", with = "alt", action = "SplitHorizontally" },
  { key = "-", with = "alt", action = "SplitVertically" },
  { key = "w", with = "control | shift", action = "ClosePane" },
  { key = "up", with = "control | shift", action = "ScrollLineUp" },
  { key = "down", with = "control | shift", action = "ScrollLineDown" },
  { key = "pageup", with = "control | shift", action = "ScrollHalfPageUp" },
  { key = "pagedown", with = "control | shift", action = "ScrollHalfPageDown" },
  { key = "home", with = "control | shift", action = "ScrollToTop" },
  { key = "end", with = "control | shift", action = "ScrollToBottom" },
]
```

inudev5:
@raphamorim  this action 'ScrollLineDown' is not defined is it?

raphamorim:
It was removed in main in the new API, now is `Scroll(1)` for `ScrollLineUp` and `Scroll(-1)` for `ScrollLineDown`

Issue #259, opened 2023-09-27
Sugarloaf: Positioning glyphs


raphamorim:
Currently Sugarloaf has a ~hacky~ solution for position the glyphs, but it can be done using `glyph_brush` (https://crates.io/crates/glyph_brush). It would get rid of bad fonts being positioned and some weird stuff rio gets as issues.

Prob the right way is drop text/section usage (https://github.com/raphamorim/rio/blob/main/sugarloaf/src/sugarloaf.rs#L438-L469) and use GlyphPositioner.

karesztrk:
Any plan on possible fixes regarding this issue?

TornaxO7:
> Any plan on possible fixes regarding this issue?

Aye: I want to tackle this.

TornaxO7:
A lot of stuff is happening in `rio` so I'll wait first.

raphamorim:
This issue should not be able to reproduce with the latest release (v0.1.0), closing this issue for now. In case find anything odd please let me know and we can reopen or create a new issue.

PR #286, opened 2023-10-14
minor suggestions on calculate_mouse_position functions.

inudev5:
its a minor refactoring suggestion to reduce function arguments.

### DIFF
--- a/rio/src/screen/mod.rs
+++ b/rio/src/screen/mod.rs
@@ -224,18 +224,7 @@ impl Screen {
 
     #[inline]
     pub fn mouse_position(&self, display_offset: usize) -> Pos {
-        calculate_mouse_position(
-            &self.mouse,
-            display_offset,
-            self.sugarloaf.layout.scale_factor,
-            (self.sugarloaf.layout.columns, self.sugarloaf.layout.lines),
-            self.sugarloaf.layout.margin.x,
-            self.sugarloaf.layout.margin.top_y,
-            (
-                self.sugarloaf.layout.scaled_sugarwidth,
-                self.sugarloaf.layout.scaled_sugarheight,
-            ),
-        )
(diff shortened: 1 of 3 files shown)


raphamorim:
Hey there @inudev5 thank you for the PR,

One of the reasons I implemented the way it was before was to avoid add `.clone()` to a function that's called/stacked a lot of times in a short span of interaction and also don't rely on sugarloaf layout contract.

Other point is that I want to get rid of clone usage in sugarloaf in the future, in general, using the clone method in Rust can be expensive in terms of performance because it creates a new instance of a value and copies all of its data. Rust has a few optimizations that can help reduce the cost of using the clone method but is good to avoid.

inudev5:
@raphamorim  thank you for your answer! it helped me understand quite more.

raphamorim:
🙏 I super appreciate the PR though!

Btw if you are looking for stuff to contribute on Rio (saw your comment on ligatures), I think these two issues are more welcoming to first time contribution:

- https://github.com/raphamorim/rio/issues/278
- https://github.com/raphamorim/rio/issues/267
- https://github.com/raphamorim/rio/issues/259

The third specially would be a huge contribution for Rio since the icons are rendered without any positioning.

inudev5:
> 🙏 I super appreciate the PR though!
> 
> Btw if you are looking for stuff to contribute on Rio (saw your comment on ligatures), I think these three issues are more welcoming to first time contribution:
> 
> * [Scroll to fast in vim and nano #278](https://github.com/raphamorim/rio/issues/278)
> * [[Keybinding] ScrollLineDown/ScrollLineUp actions not work as expected #267](https://github.com/raphamorim/rio/issues/267)
> * [Sugarloaf: Positioning glyphs #259](https://github.com/raphamorim/rio/issues/259)
> 
> The third specially would be a huge contribution for Rio since the icons are rendered without any positioning.

I see. thanks.

PR #287, opened 2023-10-14
format: reformat list to matrix

TornaxO7:
makes it, in my opinion a bit better to read.


raphamorim:
Hey there @TornaxO7 thank you for the PR,

While I do agree is better to read as a matrix, as the code you suggested was before like that, one of Rio guidelines is to drop usage of `#[rustfmt::skip]` or any removal of clippy rule completely, so Rio fully relies on rust-fmt and rust-clippy. In my experience with OSS projects, it tends to get more and more of skips/disable rules and often leads to blocks of code (including parts of code that could be improved) that aren't processed by tools like fmt/clippy.

raphamorim:
in another note maybe I should add this guideline to some document like contributing

PR #288, opened 2023-10-14
feat: added scroll_modifier to Config with default 2.0

inudev5:
added scroll_modifier to Config and set default as 2.0 
related to: #278

### DIFF
--- a/rio-config/src/defaults.rs
+++ b/rio-config/src/defaults.rs
@@ -80,6 +80,10 @@ pub fn default_window_height() -> i32 {
     400
 }
 
+pub fn default_scroll_multiplier() -> f64 {
+    2.0
+}
+
 pub fn default_config_file_content() -> String {
     r#"
 # Cursor
--- a/rio-config/src/defaults.rs
+++ b/rio-config/src/defaults.rs
@@ -81,7 +81,7 @@ pub fn default_window_height() -> i32 {
 }
 
 pub fn default_scroll_multiplier() -> f64 {
-    2.0
+    3.0
 }
 
 pub fn default_config_file_content() -> String {
--- a/rio-config/src/lib.rs
+++ b/rio-config/src/lib.rs
@@ -102,6 +102,8 @@ pub struct Config {
     pub padding_x: f32,
     #[serde(default = "default_cursor")]
     pub cursor: char,
+    #[serde(default = "default_scroll_multiplier", rename = "scroll-multiplier")]
+    pub scroll_multiplier: f64,
     #[serde(default = "default_env_vars", rename = "env-vars")]
     pub env_vars: Vec<String>,
     #[serde(default = "default_option_as_alt", rename = "option-as-alt")]
@@ -346,6 +348,7 @@ impl Default for Config {
             bindings: Bindings::default(),
             colors: Colors::default(),
             cursor: default_cursor(),
+            scroll_multiplier: default_scroll_multiplier(),
             developer: Developer::default(),
             disable_unfocused_render: false,
             env_vars: default_env_vars(),
--- a/rio/src/screen/mod.rs
+++ b/rio/src/screen/mod.rs
@@ -195,7 +195,7 @@ impl Screen {
             context_manager,
             ime,
             sugarloaf,
-            mouse: Mouse::default(),
+            mouse: Mouse::new(config),
             state,
             bindings,
             clipboard,
@@ -308,6 +308,8 @@ impl Screen {
             terminal.default_cursor_shape = cursor;
             terminal.blinking_cursor = config.blinking_cursor;
         }
+
+        self.mouse.update_multiplier(config);
 
         let width = self.sugarloaf.layout.width_u32 as u16;
         let height = self.sugarloaf.layout.height_u32 as u16;
--- a/rio/src/screen/mod.rs
+++ b/rio/src/screen/mod.rs
@@ -195,7 +195,7 @@ impl Screen {
             context_manager,
             ime,
             sugarloaf,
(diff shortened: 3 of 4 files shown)


rio-config/src/defaults.rs
```diff
@@ -80,6 +80,10 @@ pub fn default_window_height() -> i32 {
     400
 }
 
+pub fn default_scroll_multiplier() -> f64 {
+    2.0
```
raphamorim, line 84: Any reason to use `2.0` the default was 3.0 no?

rio/src/screen/mouse.rs
```diff
@@ -53,6 +54,18 @@ impl Default for Mouse {
     }
 }
 
+impl Mouse {
+    pub fn new(config: &Rc<rio_config::Config>) -> Self {
```
raphamorim, line 58: since only multiplier is configurable and the rest of the values have a default, is better to just receive multiplier instead of the reference counter reference of config.

```rust
pub fn new(multiplier: f64) -> Self {
```

rio/src/screen/mouse.rs
```diff
@@ -53,6 +54,18 @@ impl Default for Mouse {
...
+        Self {
+            multiplier: config.scroll_multiplier,
+            ..Default::default()
+        }
+    }
+    pub fn update_multiplier(&mut self, config: &Rc<rio_config::Config>) {
```
raphamorim, line 64: please rename to set_multiplier to keep consistency with other parts of rio, like https://github.com/raphamorim/rio/blob/d6b73d4a2fc2cfcf91160dfc883c8f48174fe3d4/rio/src/ime.rs#L17 , https://github.com/raphamorim/rio/blob/d6b73d4a2fc2cfcf91160dfc883c8f48174fe3d4/sugarloaf/src/font/loader/mod.rs#L420 , https://github.com/raphamorim/rio/blob/d6b73d4a2fc2cfcf91160dfc883c8f48174fe3d4/sugarloaf/src/core.rs#L43

rio-config/src/lib.rs
```diff
@@ -102,6 +102,8 @@ pub struct Config {
     pub padding_x: f32,
     #[serde(default = "default_cursor")]
     pub cursor: char,
+    #[serde(default = "default_scroll_multiplier", rename = "scroll-multiplier")]
```
raphamorim, line 105: 👍

rio-config/src/defaults.rs
raphamorim: I think is better set it back to 3.0 to ensure have less breaking changes for this release

inudev5: rolled it back. f9d755f41fbfa3d16fdfbd7a25506f0b099b7cfb

rio/src/screen/mouse.rs
inudev5: editted .  f9d755f41fbfa3d16fdfbd7a25506f0b099b7cfb

rio/src/screen/mouse.rs
inudev5: changed the name. f9d755f41fbfa3d16fdfbd7a25506f0b099b7cfb. should we call it in Screen::update_config?

rio/src/screen/mouse.rs
```diff
@@ -3,6 +3,7 @@ use crate::crosswords::pos::Line;
 use crate::crosswords::pos::Side;
 use crate::event::ClickState;
 use crate::screen::Pos;
+use std::rc::Rc;
```
raphamorim, line 6: I think this import isn't necessary anymore.

rio/src/screen/mouse.rs
```diff
@@ -53,6 +54,18 @@ impl Default for Mouse {
...
 }
 
+impl Mouse {
+    pub fn new(multiplier: f64) -> Self {
+        Self {
+            multiplier,
+            ..Default::default()
+        }
+    }
+    pub fn set_multiplier(&mut self, multiplier: f64) {
```
raphamorim, line 64: This case we can use inline,
suggestion:

```rust
#[inline]
pub fn set_multiplier(&mut self, multiplier: f64) {
```

raphamorim:
@inudev5 looks good! just left 2 comments 👌

rio/src/screen/mouse.rs
inudev5: deleted. 38271fcee8cfa52a897150708224820cc9f19a09

rio/src/screen/mouse.rs
inudev5: edited. 38271fcee8cfa52a897150708224820cc9f19a09

raphamorim (approved):
👌

rio/src/screen/mouse.rs
raphamorim: I think `update_config` should actually rename to other to something else tbh. The prefix `set_` is the default for get/set in rio but update_config does more than a regular setter, so idk (in the future I plan to revisit update_config though)

rio/src/screen/mod.rs
```diff
@@ -309,6 +309,8 @@ impl Screen {
             terminal.blinking_cursor = config.blinking_cursor;
         }
 
+        self.mouse.set_multiplier(config.scroll_multiplier);
+
```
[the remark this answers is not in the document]
raphamorim: Is correct, sorry forgot to comment.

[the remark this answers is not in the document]
inudev5: @raphamorim No worries. I appreciate your reviews. Thank you.

PR #291, opened 2023-10-16
docs: uniform style in configuration sample

ralgozino:
Use the same style in the explanations inside the configuration sample.

Add missing `:` after `Example` and remove the indentation of the example itself so when uncommenting the example code is already usable.

### DIFF
--- a/docs/docs/documentation/configuration-file.md
+++ b/docs/docs/documentation/configuration-file.md
@@ -7,7 +7,7 @@ The configuration should be the following paths otherwise Rio will use the defau
 
 MacOS and Linux configuration file path is `~/.config/rio/config.toml`.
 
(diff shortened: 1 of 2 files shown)


docs/docs/documentation/configuration-file.md
```diff
@@ -188,7 +188,7 @@ performance = "High"
 #
 # Example 1 using fish shell from bin path:
 #
-#   shell = { program = "/bin/fish", args = ["--login"] }
+#  shell = { program = "/bin/fish", args = ["--login"] }
```
raphamorim, line 191: maybe this one needs to delete just one more space?

ralgozino: good catch. I should also update the `rio-config/src/defaults.rs` file accordingly, right?

ralgozino: done ✅

Issue #311, opened 2023-10-25
Display the shell name on the tab title


eduronqui:
It would be nice to have the terminal shell name as part of the tab title.

raphamorim:
Closed in #312

PR #312, opened 2023-10-25
Add program (shell) name to the window_title

eduronqui:
Following up on #311 this PR is changing how the tab and window titles are composed to add the shell name in parenthesis on the title as per the following image.

image


raphamorim (approved):
🥇

raphamorim:
Closes #311

PR #332, opened 2023-11-06
Add web font

nix6839:
Add web font to facilitate design ease and provide a consistent experience to users.

For now, I have set it to [Noto Sans](https://fonts.google.com/noto/specimen/Noto+Sans). Do you have a preferred font?

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -21,6 +21,31 @@ const config = {
     locales: ['en'],
   },
 
+  headTags: [
+    {
+      tagName: 'link',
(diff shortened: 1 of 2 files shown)


raphamorim (approved):

raphamorim:
> For now, I have set it to [Noto Sans](https://fonts.google.com/noto/specimen/Noto+Sans). Do you have a preferred font?

No, I don't have. Rio uses by default https://github.com/microsoft/cascadia-code but I think is more a terminal font.

Thanks @nix6839 !

Issue #62, opened 2023-05-27
Support alternative config file locations


Afluttera:
Having a new application wanting to place it's configuration data in the users home directory is rather unfortunate, and alternative locations would be appreciated. It would be preferable if the default suggested location were to change as well, but that's not strictly necessary.

For Linux, the relevant directory for the config would be `$XDG_CONFIG_HOME/rio/config.toml` if `XDG_CONFIG_HOME` set, otherwise it's conventionally `~/.config/rio/config.toml` [1]
For Windows, the directory would likely be `{Home folder}/AppData/Roaming/rio/config.toml`, though, I believe there may also be special ways you're suppose to access directories like that?
For Mac, it seems as though the suggested location would be `~/Library/Application Support/rio/rio.toml`. I don't use a Mac, however, and found this through skimming documentation.

If backwards compatibility is a priority here, then what's typically done is that an application would first check the legacy path and load a config file there if present, and only look in subsequent paths if it isn't. This can create a bit of confusion about which is the proper one, but having multiple paths like this isn't especially unusual either, as long as the order in which paths are checked is documented.

[1] Also consider looking at the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which contains the proper locations of other types of data on Linux. 

Thank you for your work on rio. I'm looking forward to its progress!

Edit: I just noticed themes are also loaded from that directory. I believe the locations outlined for the config file would also be correct for these theme files as well.

Edit: Actually, it seems as though this was implemented, but just reverted in commit https://github.com/raphamorim/rio/commit/2044c3a19546d802e640539225260306da224953? Is it intended to be kept that way?

raphamorim:
Hey there @Afluttera, thanks for bring it up and adding proper details (so gets documented for future discussions).

Yes, I really like https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html and had implemented it before as you saw in [2044c3a19546d802e640539225260306da224953](https://github.com/raphamorim/rio/commit/2044c3a19546d802e640539225260306da224953). However for Mac is quite "odd" path for anyone that works mostly with terminal or comes from Linux (as my case).

I have been wondering if instead of use XDG spec blindly for all platforms, if we can have that for Windows and Linux but for MacOs use `~/.config/rio` instead of `~/Library/Application Support/rio`.

So would be like

- Linux and MacOs: `{$HOME}/.config/rio`
[...]

PR #340, opened 2023-11-16
Use standard config directory

nix6839:
Use standard config directory defined in https://dirs.dev/
Ref [this](https://github.com/dirs-dev/dirs-rs#features):
Function name | Value on Linux/Redox | Value on Windows | Value on macOS
-- | -- | -- | --
`config_local_dir` | `Some($XDG_CONFIG_HOME)` or `Some($HOME/.config)` | `Some({FOLDERID_LocalAppData})` | `Some($HOME/Library/Application Support)`

This will be a breaking change for macOS and Linux users.
- Linux:
  `~/.config/rio/` to `~/.config/rio/` or `$XDG_CONFIG_HOME/rio/`
- macOS:
  `~/.config/rio/` to `~/Library/Application Support/rio/`

How can effectively announce breaking change?

---

It also includes bug fixes on Windows, along with some refactoring:
image_2023-11-16_09-42-14
image_2023-11-16_09-43-29
The welcome message now prints a valid path separator.

### DIFF
--- a/docs/docs/configuration-file.md
+++ b/docs/docs/configuration-file.md
@@ -5,9 +5,9 @@ language: 'en'
 
 The configuration should be the following paths otherwise Rio will use the default configuration.
 
-MacOS and Linux configuration file path is `~/.config/rio/config.toml`.
-
-Windows configuration file path is `C:\Users\USER\AppData\Local\rio\config.toml` (replace "USER" with your user name).
+- macOS: `~/Library/Application Support/rio/config.toml`
+- Linux: `$XDG_CONFIG_HOME/rio/config.toml` or `~/.config/rio/config.toml`
+- Windows: `C:\Users\USER\AppData\Local\rio\config.toml` (replace "USER" with your user name)
 
 Updates to the configuration file automatically triggers Rio to render the terminal with the new configuration.
 
@@ -57,7 +57,8 @@ performance = "High"
 # Themes
 #
 # Rio looks for a specified theme in the themes folder
-# (macos and linux: ~/.config/rio/themes/dracula.toml)
+# (macos: ~/Library/Application Support/rio/themes/dracula.toml)
+# (linux: $XDG_CONFIG_HOME/rio/themes/dracula.toml or ~/.config/rio/themes/dracula.toml)
 # (windows: C:\Users\USER\AppData\Local\rio\themes\dracula.toml)
 #
 # Example:
@@ -249,9 +250,9 @@ performance = "High"
 
 # Colors
 #
-# Defining colors in the configuration file will override any                                                                                                                     
-# colors set in the theme if you're using a theme. The default                                                                                                                  
-# configuration is without a theme.    
+# Defining colors in the configuration file will override any
+# colors set in the theme if you're using a theme. The default
+# configuration is without a theme.
 #
 # Example:
 # [colors]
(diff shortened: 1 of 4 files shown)


raphamorim:
Hey @nix6839 thank you for your PR,

This was discussed other times #14 and #62 . But as marked in https://github.com/raphamorim/rio/issues/62#issuecomment-1566137402 I have no plans in adopt it mostly because for MacOS is quite odd configuration files are saved `~/Library/Application Support/rio/*` (I think makes even harder to use/manipulate the files as `~/.config` like aws, nvim and others use)

nix6839:
> Hey @nix6839 thank you for your PR,
> 
> This was discussed other times #14 and #62 . But as marked in [#62 (comment)](https://github.com/raphamorim/rio/issues/62#issuecomment-1566137402) I have no plans in adopt it mostly because for MacOS is quite odd configuration files are saved `~/Library/Application Support/rio/*` (I think makes even harder to use/manipulate the files as `~/.config` like aws, nvim and others use)

Hmmm As someone who likes to follow standards, the current behavior seems weird to me.

@raphamorim So how about keeping the default at `~/.cofig/rio` on macOS, but using that path if `$XDG_CONFIG_HOME` is set?

raphamorim:
> @raphamorim So how about keeping the default at ~/.cofig/rio on macOS, but using that path if $XDG_CONFIG_HOME is set?

I also considered and implemented that approach earlier but got more messy for users. Started to be confusing where the file could be coming from (and leading to GH issues with people asking where was supposed to be).

I would like to use the standard xdg but I also understand that it has a drawback for mac users that work mostly on unix side (like myself 😆)

raphamorim:
Also another note, the person that implemented xdg config path on rio for the first time, got disappointed after check that it comes from /Library/ and even asked to rollback so is a bit tricky IMO.

nix6839:
I think we need a way to set the path to the config file somehow.
Probably `XDG_CONFIG_HOME` would be the best way to do it, since other popular terminals use it ([alacritty](https://github.com/alacritty/alacritty?tab=readme-ov-file#configuration), [kitty](https://sw.kovidgoyal.net/kitty/invocation/#cmdoption-kitty-config), [wezterm](https://wezfurlong.org/wezterm/config/files.html#configuration-files)).

raphamorim:
There's also the problem that if XDG_CONFIG_HOME is changed will affect all other stuff that relies on that and for other cases does makes sense be a /Library/ path.

raphamorim:
Other take I though was to use a specific configuration file path `RIO_CONFIG` (in that case if doesn't exist then you could set as XDG_CONFIG_HOME value)

nix6839:
> There's also the problem that if XDG_CONFIG_HOME is changed will affect all other stuff that relies on that and for other cases does makes sense be a /Library/ path.

Normally, `XDG_CONFIG_HOME` would be changed intentionally by the user, so why would there be a problem?

> Other take I though was to use a specific configuration file path `RIO_CONFIG` (in that case if doesn't exist then you could set as XDG_CONFIG_HOME value)

I think that would be a good option too 😀 .

raphamorim:
> Normally, XDG_CONFIG_HOME would be changed intentionally by the user, so why would there be a problem?

A problem would be take everything together with Rio or other application.

> I think that would be a good option too 😀 .

Would you like to create an issue describing it? So we would have kinda of a plan 👌

nix6839:
> Would you like to create an issue describing it? So we would have kinda of a plan 👌

I'll try it after sleeping and waking up😁.

PR #349, opened 2023-11-24
bumping `winit`

TornaxO7:


TornaxO7:
Eh what? Isn't there a `main` function?

Cargo.toml
```diff
@@ -51,7 +51,7 @@ wasm-bindgen-test = "0.3"
 env_logger = "0.10"
 console_log = "0.2"
 console_error_panic_hook = "0.1.7"
-winit = { git = "https://github.com/raphamorim/winit", default-features = false, features = ["rwh_05"] }
+winit = { version = "0.29", default-features = false, features = ["x11", "wayland", "wayland-dlopen", "wayland-csd-adwaita", "rwh_05"] }
```
raphamorim, line 54: the features shouldn't be added by default in the workspace.

For each package is mapped the necessary features

https://github.com/raphamorim/rio/blob/main/frontends/cross-winit/Cargo.toml#L61-L72

raphamorim: Also could lock `0.29.4`? I prefer this way since it helps me to debug between rio versions

raphamorim:
```suggestion
winit = { version = "0.29.4", default-features = false, features = ["rwh_05"] }
```

TornaxO7: > Also could lock `0.29.4`? I prefer this way since it helps me to debug between rio versions

hm... I don't think that pinning the minor version will change that much but if you insist, I don't mind.

raphamorim: It doesn't but helps me to understand/memorize which version is using what (it was good to quick investigate a wpgu issue rio had before)

raphamorim (approved):
🏅

PR #350, opened 2023-11-24
renaming `rust-toolchain` to `rust-toolchain.toml`

TornaxO7:
that enables on my editor syntax highlighting per default xD

Makes it also clear which file type it has :)


TornaxO7:
> ^ consider adding a `main` function to `wa/examples/mouse_cursor.rs`

I don't quite understand why I'm getting this :thinking:

PR #351, opened 2023-11-24
small refactor suggestions

TornaxO7:
Just some small refactor suggestions which should make the code more readable (in my opinion).


frontends/cross-winit/src/main.rs
```diff
@@ -97,29 +95,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let options = cli::Options::new();
-
-    let mut config_error: Option<rio_backend::config::ConfigError> = None;
-    let mut config = match rio_backend::config::Config::try_load() {
-        Ok(config) => config,
-        Err(error) => {
-            config_error = Some(error);
-            rio_backend::config::Config::default()
-        }
+    let cli = cli::Cli::parse();
```
raphamorim, line 98: having cli as variable name and mod name is a bit confusing, can we rename this to other name? could even be `parsed_cli`, `initialized_cli` etc

TornaxO7: I changed it to `args`.

raphamorim (approved):

PR #352, opened 2023-11-24
Fix logical keybinding detection.

TornaxO7:
Rio is reading the correct logical key after removing `key.key_without_modifiers()` which fixes my issue of not getting my keys of being registered.


TornaxO7:
Closed in favor of https://github.com/raphamorim/rio/commit/3b0b9fa8b5595ae8926064ce1cbba1a8978db3ca

PR #368, opened 2023-12-06
feat(i18n): add spanish translation to landing page

ralgozino:
This PR adds the Spanish translation to the landing page :) 

There are still some texts that are not translated, like the footer and the page title; but this is common to the other translations (because these seem to be "hard coded"), so I figured that I should open the PR with what I have until we can translate the rest.


raphamorim (approved):
🥇 that's great!

Issue #344, opened 2023-11-19
Kitty-keyboard-protocol causes Backspace to delete 2 characters.


gyreas:
Hi, 
I'm trying out Rio first time, so after installation, I enabled some fancy-looking options, after which backspace begins to delete 2 characters instead of one. By process of elimination, `use-kitty-keyboard-protocol = true` turned out to be the culprit, without which the default value `false` works well.

To reproduce:
1. Create a temp config with cursor, blinking-cursor, use-kitty-keyboard-protocol=true and performance="high" options set
2. Open a new Rio with that config
3. Type some random characters on the cmdline and press backspace once
Observation: **Two characters before the cursor are deleted.**
Expected Outcome:  **The character before the cursor is deleted.**

raphamorim:
@gyreas thanks for the issue, are you using Rio in which version and platform?

raphamorim:
Should be fixed after @hougesen fix, let me know anything

PR #375, opened 2023-12-10
fix: remove duplicate kitty backspace keybinds

hougesen:
The deleted keybindings were duplicates of https://github.com/raphamorim/rio/blob/208e913ed727c19e61847638f4ff226e37a0a189/frontends/cross-winit/src/screen/bindings/mod.rs#L742-L744 with the only difference between them being the spaces.

After deleting the lines I no longer had any issues with backspace deleting 2 characters when Kitty protocol is turned on (#344).

Before with Kitty `false`:
Before with Kitty false

Before with Kitty `true`:
Before with Kitty true

After with Kitty `false`:

After with Kitty false

After with Kitty `true`:

After with Kitty true


raphamorim (approved):
Good finding! I forgot to remove it when porting kitty keyboard to config.
Tested and looks good 👌

PR #378, opened 2023-12-11
fix: update padding top on config change

hougesen:
Currently the padding top is checked on startup, but not when the config is changed, which means going from `BottomTab` to `TopTab` will result in the navigation overlapping the first line (left @ f26f633):

BottomTab to TopTab

`TopTab` to `BottomTab` results in a "wasted" line:

TopTab to BottomTab

This should fix it :)


raphamorim (approved):
Thanks @hougesen

PR #381, opened 2023-12-12
refactor(cross-winit): remove unnecessary cloning of key events

hougesen:
I know it is pretty minor, but while jumping around in the code I noticed a few places cloning could be removed. 

Feel free to close the PR if I am mistaken, or there is a good reason for it :)


raphamorim (approved):

raphamorim:
> I know it is pretty minor, but while jumping around in the code I noticed a few places cloning could be removed.

Please keep it come haha less cloning the better, there's a lot of cloning code that became a debt in the source code.

PR #391, opened 2023-12-17
docs: reordering keys to avoid conflicts

AntonyZ89:
Reordering non-grouped keys to the top to prevent conflicts and unexpected scenarios.

Reason for this change:

In my OS (Arch Linux + Hyprland WM), the shell key is being merged with the [fonts] properties.

My config.toml:

```toml
# some docs here
[fonts]
family = "Maple Mono SC NF"
size = 20 

# some docs here
shell = { program = "/usr/bin/tmux", args = ["new-session", "-c", "/var/www"] }
```

As a result, an interpretation of how `rio-backend` sees the "fonts" group:
```json
{
  "fonts": {
    "family": "Maple Mono SC NF",
    "size": 20,
    "shell": {
       "program": "/usr/bin/tmux",
       "args": ["new-session", "-c", "/var/www"]
    }
  }
}
```
and shell is None because it was merged into fonts, disregarding my configuration.

My suggestion is to move up all keys that aren't grouped with brackets ([example]) to avoid merging and unexpected scenarios:

```toml
# non-grouped keys first
cursor = '▇'
theme = "catppuccin-macchiato"
padding-x = 0
env-vars = []
use-fork = false

# some doc here
[fonts]
family = "Maple Mono SC NF"
size = 20 

# some doc here
[shell]
program = "/usr/bin/tmux"
args = ["new-session", "-c", "/var/www"] 
```

### DIFF
--- a/docs/docs/configuration-file.md
+++ b/docs/docs/configuration-file.md
@@ -90,6 +90,35 @@ Possible choices: `both`, `left` and `right`.
 option-as-alt = 'left'
 ```
 
+## Startup directory
+
+Directory the shell is started in. If this is unset the working directory of the parent process will be used.
+
+This configuration only works if `use-fork` is disabled
+
+```toml
+working-dir = "/Users/raphael/Documents/"
+```
+
+## Environment variables
+
+The example below sets fish as the default SHELL using env vars, please do not copy this if you do not need it.
+
+```toml
+env-vars = []
+```
+
+## Use fork
+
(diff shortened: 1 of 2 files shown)


raphamorim:
> My suggestion is to move up all keys that aren't grouped with brackets ([example]) to avoid merging and unexpected scenarios:

Yes, I agree. Thanks for the PR @AntonyZ89

raphamorim (approved):

PR #396, opened 2023-12-19
docs: set padding-x default to 0

hougesen:
`padding-x` was changed to 0 on all platforms in `v0.0.30`. 

[https://github.com/raphamorim/rio/blob/1fa31f7cf6caa1c0c4caf2a5268eb7cea8799d6a/rio-backend/src/config/defaults.rs#L6-L9](https://github.com/raphamorim/rio/blob/1fa31f7cf6caa1c0c4caf2a5268eb7cea8799d6a/rio-backend/src/config/defaults.rs#L6-L9)

### DIFF
--- a/docs/docs/configuration-file.md
+++ b/docs/docs/configuration-file.md
@@ -74,10 +74,10 @@ theme = "dracula"
 
 ## Padding-x
 
-Define x axis padding (default is 10)
+Define x axis padding (default is 0)
 
 ```toml
-padding-x = 10
+padding-x = 0
 ```
 
 ## Option as Alt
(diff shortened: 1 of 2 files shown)


docs/docs/configuration-file.md
```diff
@@ -74,10 +74,10 @@ theme = "dracula"
 
 ## Padding-x
 
-Define x axis padding (default is 10)
+Define x axis padding (default is 0)
 
 ```toml
-padding-x = 10
```
raphamorim, line 80: I would like to keep the bottom example (so people can have an idea how the updates are done) but the `(default is 10)` needs to be fixed to `0`

hougesen: Makes sense, i'll fix it

raphamorim (approved):
Thanks @hougesen ! 🙏

PR #397, opened 2023-12-20
docs: use-current-path requires use-fork to be false

hougesen:
`create_pty_with_fork` does not support starting in specific path (*yet?*).

[https://github.com/raphamorim/rio/blob/f4f8d6c381d905c7c108efacf6132ea0c9614aaf/teletypewriter/src/unix/mod.rs#L581](https://github.com/raphamorim/rio/blob/f4f8d6c381d905c7c108efacf6132ea0c9614aaf/teletypewriter/src/unix/mod.rs#L581)

### DIFF
--- a/docs/docs/configuration-file.md
+++ b/docs/docs/configuration-file.md
@@ -308,7 +308,7 @@ divider = 1.0
 	- `Plain`
 
 - `clickable` - Enable click on tabs to switch.
-- `use-current-path` - Use same path whenever a new tab is created.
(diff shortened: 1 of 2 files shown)


RaySlash:
hey, you should probably push to a different branch as mentioned here:
https://github.com/raphamorim/rio/blob/fb689b7e1cfe3d8b4266f569f139c1fa00fa7d68/README.md?plain=1#L5

raphamorim:
> hey, you should probably push to a different branch as mentioned here:

Yes, rio-backend needed to be in the 0.0.x branch but no worries I can cherry-pick this commit to there. Also I need to fix the PRs to use the CI of 0.0.x

raphamorim:
Thanks @hougesen 🙏

Issue #356, opened 2023-11-27
Characters are missing


TornaxO7:
![image](https://github.com/raphamorim/rio/assets/50843046/9b20ff27-6cb5-432e-96e2-bdaee65d4213)

On the left is `rio` and on the right is `alacritty`. As you can see `rio` isn't displaying everything.

![image](https://github.com/raphamorim/rio/assets/50843046/669f2a1f-e84f-4eca-9871-b3d03c45af56)

also characters disappear if I press `tab` to get tab-completion.

TornaxO7:
![image](https://github.com/raphamorim/rio/assets/50843046/857f6ee5-d35e-4110-a843-b789885e1532)

also happening with
```
rioterm 0.0.28
```

PR #399, opened 2023-12-25
Fix missing characters when rendering

TornaxO7:
Closes https://github.com/raphamorim/rio/issues/356

So that's how it looks like currently with the latest release of `rio`:

image

That's how it looks like *at the moment* with the PR:

image

I don't really understand, why this `2. *` multiplicator fixes this. According to my debbuging the rows are correctly read, so the `stack` is correct but afterwards... I'm confused why `+ 1.` isn't enough for the `sugar_width`. May I ask if you have an idea?

**Note:**
Yes, currently the `flake` branch is added here as well, because I need it to tinker with `rio` xD

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3642,9 +3642,9 @@ checksum = "ed94fce61571a4006852b7389a063ab983c02eb1bb37b47f8272ce92d06d9538"
 
 [[package]]
 name = "winit"
-version = "0.29.5"
+version = "0.29.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2cc935117ef48caed8822a893efef723e07ad868f668dfc4d244aea8873b07f9"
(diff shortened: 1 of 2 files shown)


raphamorim:
>  Yes, currently the flake branch is added here as well, because I need it to tinker with rio xD

I see, let's wait your PR with nix be merged then, is just small comments.

> I don't really understand, why this 2. * multiplicator fixes this. According to my debbuging the rows are correctly read, so the stack is correct but afterwards... I'm confused why + 1. isn't enough for the sugar_width. May I ask if you have an idea?

I don't think the fix should be increase the `width_bound`, if the width_bound is increased then is better to remove. What this property does is ensure that a character with width 1 or 2 would never pass the cell bound (for example cell width 1 with font bound 14px can never exceed that). If this property isn't working it means either the font bound is being computed wrong (we need to check) or other property is being inputted incorrect.

PR #412, opened 2024-01-13
flake: add `apps` entry and fix runtime error

TornaxO7:
Fixes the core dump when running `rio` with `nix`. @RaySlash you can take a look if you want.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,136 +11,56 @@
     };
   };
 
-  outputs = inputs @ {flake-parts, ...}:
-    flake-parts.lib.mkFlake {inherit inputs;} {
-      imports = [];
+  outputs = inputs @ { flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [ ];
 
       systems = import inputs.systems;
 
-      perSystem = {
-        config,
-        self',
-        inputs',
-        pkgs,
-        system,
-        lib,
-        ...
-      }: let
-        cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
-        rust-toolchain = (pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml).override {
-          extensions = ["rust-src" "rust-analyzer"];
-        };
+      perSystem =
+        { config
+        , self'
+        , inputs'
+        , pkgs
+        , system
+        , lib
+        , ...
+        }:
+        let
+          rust-toolchain = (pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml).override {
+            extensions = [ "rust-src" "rust-analyzer" ];
+          };
 
-        mkRio = {
-          rustPlatform,
-          stdenv,
-          lib,
-          fontconfig,
-          darwin,
-          gcc-unwrapped,
-          libGL,
-          libxkbcommon,
-          vulkan-loader,
-          libX11,
-          libXcursor,
-          libXi,
-          libXrandr,
-          libxcb,
-          wayland,
-          ncurses,
-          pkg-config,
-          cmake,
-          autoPatchelfHook,
-          withX11 ? !stdenv.isDarwin,
-          withWayland ? !stdenv.isDarwin,
-          ...
-        }: let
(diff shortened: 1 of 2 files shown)


TornaxO7:
Why is the windows build failing???

flake.nix
```diff
@@ -102,7 +132,11 @@
         packages.default = self'.packages.rio;
         devShells.default = self'.devShells.msrv;
 
-        packages.rio = rustPackage "rio";
+        apps.default = {
+          type = "app";
+          program = self'.packages.rio;
```
RaySlash, line 137:
```suggestion
          program = self'.packages.default;
```
This would be better to point to rather than `packages.rio` for just the sake of keeping it logical and clean.

flake.nix
```diff
@@ -31,50 +31,78 @@
...
-          ]
-          ++ lib.optionals stdenv.isLinux [
-            pkg-config
-            cmake
-            autoPatchelfHook
-          ];
-
-        rustPackage = rust-toolchain:
-          pkgs.rustPlatform.buildRustPackage {
+        mkRio = {
```
RaySlash, line 34: This seems like an old snippet from old config? What was the motivation for this? However, I just tested this change and seems working on this end. Is this change necessary? I believe that we should not be adding additional functions on functions just for the dependencies. We could add these for `PerSystem` and see if that would work for us. I did this way of passing function carrying evaluations of dependencies to the function defined by `rustPlatform` as an experiment mostly. I am not certain if this could cause further breakage. Also, if this is for people to use flakes to generate a package for their config should not be really a concern as `nixpkgs` is being constantly updated by you and other maintainers as I can see now. I believe `nix develop` and `nix build` should be our focus. I would like some clarity regarding this.

flake.nix
RaySlash, line 137:
```suggestion
          program = self'.packages.default;
```
This would be better to point to rather than `packages.rio` for just the sake of keeping it logical and clean.

RaySlash (changes requested):

flake.nix
RaySlash: Also, just adding 
```nix
pkgs.rustPlatform.buildRustPackage {
   inherit (cargoToml.workspace.package) version;
   name = "rio";
   src = ./.;
   cargoLock.lockFile = ./Cargo.lock;
   
   cargoBuildFlags = "-p rioterm";
   
   buildInputs = runtimeDeps ++ buildDeps;
   nativeBuildInputs = buildDeps;
   runtimeDependencies = runtimeDeps;
   
   buildNoDefaultFeatures = true;
   buildFeatures = ["x11" "wayland"];
   
   meta = {
            description = "A hardware-accelerated GPU terminal emulator focusing to run in desktops and browsers";
            homepage = "https://raphamorim.io/rio";
            license = lib.licenses.mit;
            platforms = lib.platforms.unix;
            changelog = "https://github.com/raphamorim/rio/blob/master/CHANGELOG.md";
            mainProgram = "rio";
    };
};
```
was not enough to get rid of the issue where it core dumps with `NoWaylandLib` or `NoXCBLib`.

TornaxO7: > This seems like an old snippet from old config? What was the motivation for this? (...) Is this change necessary?

I'm not fully understanding this yet as well, but I assume that working with `pkgs` instead of an isolating function, as with `mkRio` here, potentially breaks some things, since `callPackage` does some "black magic" (that's what I got told of from the unofficial nixos-discord server).

However, [I asked on the discourse](https://discourse.nixos.org/t/correct-way-for-adding-dependencies-to-buildrustpackage/38219) what the preferred way is to declare a package and it seems to be either to move that into an extra file, [as suggested in the `nix pills`](https://nixos.org/guides/nix-pills/callpackage-design-pattern#id1426) (I only have skimmed through it at the time of writing this) or moving this into a variable, as we do here with `mkRio`.

> I believe nix develop and nix build should be our focus.

I agree, but I'd say that `nix run` is more important than `nix build` because what's the point of having a succeeding build, if you can't use it? ;)

raphamorim (approved):

RaySlash (approved):
LGTM

PR #420, opened 2024-01-17
Use cargo-nextest for running tests

orhun:
faster terminal deserves faster running tests thus faster CI

### DIFF
--- a/.github/workflows/test-beta.yml
+++ b/.github/workflows/test-beta.yml
@@ -24,4 +24,7 @@ jobs:
       - run: rustup component add rustfmt clippy
       - run: cargo fetch
       - run: cargo clippy -p rioterm-beta -- -D warnings
-      - run: cargo test -p rioterm-beta
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
+      - name: Run tests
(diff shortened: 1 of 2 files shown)


orhun:
what an honor

.github/workflows/test-beta.yml
```diff
@@ -24,4 +24,7 @@ jobs:
       - run: rustup component add rustfmt clippy
       - run: cargo fetch
       - run: cargo clippy -p rioterm-beta -- -D warnings
-      - run: cargo test -p rioterm-beta
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
```
hougesen, line 28: Adding nextest as a dev dependency would allow it to be cached by the `Swatinem/rust-cache` action, increasing the speed even further 🚀  (I think) .

orhun:
Looks like macos has a major slow down with this - some reports that I found:
- https://github.com/nextest-rs/nextest/issues/1161
- https://github.com/nextest-rs/nextest/issues/262

Feel free to re-consider this when `cargo-nextest` performance is stabilized.

orhun:
image

PR #430, opened 2024-01-22
docs: add keybind mode & text section

hougesen:
Added sections about keybind `mode` and `text`. 

Also changed the default keybind sections from bullet point to table, since it makes it easier to read.


raphamorim (approved):
Much better! 👌

Issue #279, opened 2023-10-10
Cursor shape isn't restored


GaspardCulis:
### Issue 

If I change my cursor shape to a pipe for example
```toml
# ~/.config/rio/config.toml
cursor = '|'
```
And then launch a process that changes the cursor's shape like a CLI text editor (vim or helix), when I exit the process the cursor shape does not get restored to the one defined in config.toml.

### Expected behavior

The cursor shape should get restored to the one defined in the config.

Something like xfce4-terminal has the expected behavior.

#### Rio:
```
❯ rio --version
rioterm 0.0.22
```

#### OS:

```
❯ cat /etc/os-release
NAME="EndeavourOS"
PRETTY_NAME="EndeavourOS"
ID="endeavouros"
ID_LIKE="arch btw"
BUILD_ID="2023.05.28"
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://endeavouros.com"
DOCUMENTATION_URL="https://discovery.endeavouros.com"
SUPPORT_URL="https://forum.endeavouros.com"
BUG_REPORT_URL="https://forum.endeavouros.com/c/arch-based-related-questions/bug-reports"
PRIVACY_POLICY_URL="https://endeavouros.com/privacy-policy-2"
LOGO="endeavouros"
❯ uname -r
6.5.6-zen2-1-zen
```

raphamorim:
Hey @GaspardCulis thanks for the issue,

It always lead to back to Block right? (It is hardcoded in the term logic as the default) will mark to take a look.

raphamorim:
Fixed in the main

![fix-cursor](https://github.com/raphamorim/rio/assets/3630346/0b2c090b-46a8-4276-9e67-5fda6cc25464)

vinoff:
I am having this issue:

Steps to reproduce:

- open vim
- enter insert mode (via `i` or `a`)
- cursor does not change to input mode cursor. It stays a block when it should be a vertical line.

If you move the cursor around, it changes to the vertical line.

The opposite also happens, that is, if you are in insert mode and switch to normal mode, the cursor does not change to block until you move the cursor around.


Interestingly, if you enter insert mode via "`o`", the cursor behaves correctly. Probably because it changes position automatically.

I am using rioterm 0.2.2.

raphamorim:
Hey @vinoff could you create a different issue? (using same text)

I feat this issue context has become too long for me to even keep track 🙏

PR #433, opened 2024-01-25
fix(crosswords): show beam cursor in nvim insert mode when using Zellij

hougesen:
I recently switched to Zellij, but I was having some trouble getting the cursor to change from `▇` to `|` when going into insert mode, in NeoVim.

image

Removing the following lines resulted in the cursor changing as expected:

https://github.com/raphamorim/rio/blob/0cbed145e864d0e0d6c724b09f5a41fbd7892881/rio-backend/src/crosswords/mod.rs#L1100-L1102

Which is rather ironic since it was introduced as a fix for the cursor shape not being restored in the first place (https://github.com/raphamorim/rio/commit/e6d582775356a358346c492d1a47bb06a7a63d14).

For now I have changed the code above to

```rust
content = if content == CursorShape::Beam {
	CursorShape::Beam
} else {
	self.default_cursor_shape
};
```

which fixes the issue for me. As far as I can tell it does not reintroduce the original bug (#279), but I am unsure if this change has other unwanted side effects.

### DIFF
--- a/rio-backend/src/crosswords/mod.rs
+++ b/rio-backend/src/crosswords/mod.rs
@@ -1098,7 +1098,11 @@ impl<U: EventListener> Crosswords<U> {
 
         // If is not using app cursor then use default
         if content != CursorShape::Hidden && !self.mode.contains(Mode::APP_CURSOR) {
-            content = self.default_cursor_shape;
+            content = if content == CursorShape::Beam {
+                CursorShape::Beam
+            } else {
+                self.default_cursor_shape
+            };
         }
 
         CursorState { pos, content }


hougesen:
The following modes are enabled when in Zellij + nvim

- `SHOW_CURSOR`
- `APP_KEYPAD`
- `BRACKETED_PASTE`
- `SGR_MOUSE`
- `LINE_WRAP`
- `ALT_SCREEN`
- `MOUSE_DRAG`
- `ALTERNATE_SCROLL`
- `URGENCY_HINTS`

Which means another way to "fix" my issue would be to check whether mode is alt screen (i think):

```rust
if content != CursorShape::Hidden
	&& !self.mode.contains(Mode::APP_CURSOR)
	&& !self.mode.contains(Mode::ALT_SCREEN)
{
	content = self.default_cursor_shape;
}
```

hougesen:
Closing since it was fixed in https://github.com/raphamorim/rio/commit/413c8289e60856cd72b43dac871b7ec46419107f

Issue #437, opened 2024-01-28
Cursor blinking is triggered by changes in inactive tabs


hougesen:
Changes in other tabs should not cause the active tab to rerender since it makes the ui seem buggy.  

[Screencast from 2024-01-28 16-23-43.webm](https://github.com/raphamorim/rio/assets/56034786/46bf027b-16bd-4d04-9ac8-6f94d1702668)

hougesen:
Resizing the window also triggers a lot of rerenders.

[Screencast from 2024-01-28 16-40-10.webm](https://github.com/raphamorim/rio/assets/56034786/285dd7dc-b99e-40b1-9618-e9d8ec74d08b)

raphamorim:
Good findings, yes those behaviors should be fixed. Now cursor can come either from terminal and state, most of times comes from state, but would be good move part of cursor rendering logic outside of state (state reflects in the current window rendering state).

hougesen:
Are the changes required affected by #428?

I also noticed selecting text is a lot slower when `blinking-cursor` is turned on.

Turned off:

[Screencast from 2024-01-28 23-58-52.webm](https://github.com/raphamorim/rio/assets/56034786/8de64294-c7cd-47c9-9572-2e905e1f50bb)

Turned on:

[Screencast from 2024-01-28 23-59-41.webm](https://github.com/raphamorim/rio/assets/56034786/0ae43942-94db-4bbc-acd0-89df39990ee3)

It is not a problem when using nvim or zellij, which I am guessing is because they handle cursor state themself (?).

raphamorim:
> Are the changes required affected by https://github.com/raphamorim/rio/pull/428?

No shouldn't be affected, #428 should touch mostly sugarloaf crate.

> I also noticed selecting text is a lot slower when blinking-cursor is turned on.

Hmm, that would make sense because is stacking rendering and blinking (rendering) in the same time.

> Adding a check for whether there is a selection eliminates the delay after first render.

I like the solution for now but shouldn't be the final fix (I guess) because could be that users would expect the cursor to blink if have selected text (?? not sure tbh).

Would you mind create a PR with those changes?

hougesen:
Sure, coming up.

hougesen:
Sadly this was not fixed as a by product of `v0.1.0` 😭

raphamorim:
@hougesen is this happening still right? I've tried to reproduce locally without success. Is your configuration for cursor blinking true?

raphamorim:
[...]

PR #441, opened 2024-01-29
refactor: disable cursor blink on selection (ref #437)

hougesen:
Disables `blinking-cursor` when selecting text, since it results in the application scheduling a lot of “unnecessary” rerenders, as described in #437.

There is still a small delay on first selection render, but it is not noticeable after.

_With the change_:

[Screencast from 2024-01-30 00-20-41.webm](https://github.com/raphamorim/rio/assets/56034786/468338b2-fdf6-498e-af8b-8673dc2520e4)

_Without_:

[Screencast from 2024-01-28 23-59-41.webm](https://github.com/raphamorim/rio/assets/56034786/0ae43942-94db-4bbc-acd0-89df39990ee3)

### DIFF
--- a/frontends/rioterm/src/app/route.rs
+++ b/frontends/rioterm/src/app/route.rs
@@ -1530,7 +1530,11 @@ impl Route {
 
                 // In this case the configuration of blinking cursor is enabled
                 // and the terminal also have instructions of blinking enabled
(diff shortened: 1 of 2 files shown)


raphamorim (approved):
fair change, as we talked before maybe we need to reevaluate it in the future in case users want to have the blinking cursor with selection but for now is a performance issue that needs to be addressed

PR #470, opened 2024-03-06
fix: build on Darwin

lovesegfault:
This fixes the build on Darwin by adding a dependency on MetalKit. I also did
some drive-by fixes of the `envrc` setup.

- **fix(envrc): reload when pkgRio.nix changes**
- **fix(envrc): evaluate the nix shellHook**
- **fix(pkgRio): add MetalKit for build on Darwin**

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,3 @@
 use flake
+
+watch_file pkgRio.nix
--- a/pkgRio.nix
+++ b/pkgRio.nix
@@ -29,6 +29,7 @@ let
       darwin.libobjc
       darwin.apple_sdk_11_0.frameworks.AppKit
       darwin.apple_sdk_11_0.frameworks.AVFoundation
+      darwin.apple_sdk_11_0.frameworks.MetalKit
       darwin.apple_sdk_11_0.frameworks.Vision
     ]
     else


raphamorim:
Hey @lovesegfault thanks for the PR! For me the changes looks ok but will ask @TornaxO7 and @RaySlash to also take a look 🙏

.envrc
```diff
@@ -1 +1,5 @@
 use flake
+
+watch_file pkgRio.nix
+
+eval "$shellHook"
```
TornaxO7, line 5: May I ask what this does does?

TornaxO7 (changes requested):

.envrc
lovesegfault: `shellHook` is documented here: https://nixos.org/manual/nix/stable/command-ref/nix-shell#description

Basically, some derivations define that as a variable which the dev shell should evaluate to perform additional setup. For example, when using `pre-commit-hooks.nix` they use `shellHook` to configure the `pre-commit-hooks` when entering the shell.

Upon checking, it seems explicitly saying `eval "$shellHook"` is no longer needed with `use flake` as direnv now does it correctly. I'll remove it :)

TornaxO7 (approved):
peepoAwesome

.envrc
TornaxO7: > shellHook is documented here: https://nixos.org/manual/nix/stable/command-ref/nix-shell#description

Oh, I thought it was a feature from `direnv` itself so I was wondering why it was nowhere documented in `direnv`. Thank you for the explanation.

raphamorim (approved):
👌

PR #486, opened 2024-04-09
Fix `LD_LIBRARY_PATH` in Nix flake dev shell

pbrinkmeier:
The Nix flake dev shell needs to pull conditionally loaded dynamic libraries such as `libxkbcommon` into scope manually because the `patch` step is missing when using `cargo` directly. The previous implementation of the dev shell includes the `buildInputs` into `LD_LIBRARY_PATH`, which are just the headers and not the dynamic libraries. This PR fixes that issue and makes the shell work on my machine :tm: .

c/o @TornaxO7 who recommended I open this PR.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -35,11 +35,12 @@
 
           mkDevShell = rust-toolchain:
             let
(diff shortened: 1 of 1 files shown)


raphamorim (approved):
👌 thanks @pbrinkmeier

PR #505, opened 2024-05-14
Fix README.md installation links

coelhucas:
Very minor, so I ended up not even opening an issue for this, just a fix for installation links sending to old (invalid) ones.

**EDIT:** If it makes more sense, I can target this change to `0.x.x` branch instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@
 
 | Name | Details |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


raphamorim (approved):
Good catch!! Thanks for the pr!
Regarding the 0.x.x branch, sure 👌 will be good to have although the docs there are quite outdated

PR #508, opened 2024-05-23
docs: add MacPorts install info

herbygillot:
https://ports.macports.org/port/rio/


raphamorim (approved):
wow! thank you @herbygillot for porting it!

PR #523, opened 2024-06-04
Font promotion and demotion

raphamorim:


raphamorim:
This PR isn't necessary anymore since font deallocation was implemented #670

PR #572, opened 2024-07-22
Revert "nix: add hash value for a dependency"

noaccOS:
Reverts raphamorim/rio#516

Using nix to compile rio currently fails with `error: A hash was specified for dpi-0.1.1, but there is no corresponding git dependency.`

With this I'm successfully able to use `nix run`.


raphamorim (approved):

raphamorim:
Thanks @noaccOS !

PR #589, opened 2024-08-12
Fixed custom key bindings URL in default config

rdbo:
I tried entering the URL in the default config, and it showed "Page not found"
I found the actual URL, and replaced it in `default.rs`: https://raphamorim.io/rio/docs/next/key-bindings#custom-key-bindings


raphamorim (approved):
Thank you 👌

PR #593, opened 2024-08-14
Show tab numbers in CollapsedTab layout

diwasrimal:


raphamorim:
That's great, thanks for the PR @diwasrimal. I haven't tested locally yet but I want to ask if is possible to add this feature behind a configuration property?

Maybe:

```toml
[navigation]
display-tab-number = true
```

diwasrimal:
Other modes like `TopTab`, `BottomTab` and `Breadcrumb` already show tab numbers by default so I think you can just make it default to show numbering for `CollapasedTab` too. Using a single configuration variable like `display-tab-number` that handles showing number for all modes is good too. For me, this would take a lot of time since I'm not so comfortable with rust.

raphamorim:
> Using a single configuration variable like display-tab-number that handles showing number for all modes is good too

yes, that’s what I meant, sorry if I wasn’t clear 🙏 

The problem of implementing numbers for collapsed tabs by default is that many users like the way it’s (without much details) so might been conflicting with them as well.

> For me, this would take a lot of time since I'm not so comfortable with rust.

there’s no rush! Take your time and feel free to ping me on discord as well. I can help pointing in the code 👌

PR #614, opened 2024-08-25
feat(tabs): add command for closing all tabs except the current one

### LINKED ISSUE #608: Close all tabs except for the curent one
Close all tabs except for the curent one

orhun:
closes #608

```toml
[bindings]
keys = [
  { key = "q", with = "control", action = "closeunfocusedtabs" },
]
```

Tested on Linux, works well except that it does not kill the processes on the closed tabs. I tried the following snippet to do that:

```rust
self.contexts.retain(|ctx| {
    let keep_ctx = ctx.route_id == current_route_id;
    #[cfg(not(target_os = "windows"))]
    {
        if !keep_ctx {
            let pid = ctx.shell_pid;
            if pid > 0 {
                teletypewriter::kill_pid(pid as i32);
            }
        }
    }
    keep_ctx
});
```

But then the whole terminal process is being killed. Not sure what's going on there.

Also, I was not sure what to do with the wasi stuff (what even is that?) so I simply duplicated the logic there. Let me know if it is appropriate.

### DIFF
--- a/docs/docs/key-bindings.md
+++ b/docs/docs/key-bindings.md
@@ -87,19 +87,19 @@ language: 'en'
 
 ### Search
 
-| Action               | Key                                                         |
-| -------------------- | ----------------------------------------------------------- |
-| Search forward         | `Command + f`          |
-| Search backward        | `Command + b`          |
-| Quit search            | `Escape` (while search is open)        |
-| Quit search            | `Control + c` (while search is open)   |
-| Focus next             | `Enter` (while search is open)         |
-| Focus previous         | `Shift + Enter` (while search is open) |
-| Clear search word      | `Control + w` (while search is open)   |
-| Search history previous| `Control + p` (while search is open)   |
-| Search history previous| `ArrowUp` (while search is open)       |
-| Search history next    | `Control + n` (while search is open)   |
-| Search history next    | `ArrowDown` (while search is open)     |
+| Action                  | Key                                    |
+| ----------------------- | -------------------------------------- |
+| Search forward          | `Command + f` / `Control + Shift + f`  |
+| Search backward         | `Command + b` / `Control + Shift + b`  |
+| Quit search             | `Escape` (while search is open)        |
+| Quit search             | `Control + c` (while search is open)   |
+| Focus next              | `Enter` (while search is open)         |
+| Focus previous          | `Shift + Enter` (while search is open) |
+| Clear search word       | `Control + w` (while search is open)   |
+| Search history previous | `Control + p` (while search is open)   |
+| Search history previous | `ArrowUp` (while search is open)       |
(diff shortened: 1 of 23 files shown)


docs/docs/key-bindings.md
```diff
@@ -190,6 +190,7 @@ Execute a predefined action in Rio terminal.
 | :------------------- | :------------------------------------------------------------------ |
 | CreateTab            |                                                                     |
 | CloseTab             |                                                                     |
+| CloseOtherTabs       |                                                                     |
```
raphamorim, line 193: Can we rename it for `CloseUnfocusedTabs`?

raphamorim:
> But then the whole terminal process is being killed. Not sure what's going on there.

Hmm, I take a look but I guess you can kill the process but cannot remove the context while doing it. Will check

raphamorim:
> Also, I was not sure what to do with the wasi stuff (what even is that?) so I simply duplicated the logic there. Let me know if it is appropriate.

WA is a replace in development for winit, for now you can ignore it 😄

raphamorim:
Interesting! For me it does work locally

demo-close

raphamorim:
just didn't check if killed the ttys though

raphamorim:
@orhun could you try with

```rust
#[inline]
    pub fn close_unfocused_tabs(&mut self) {
        let current_route_id = self.current().route_id;
        self.titles.titles.retain(|&i, _| i == self.current_index);

        #[cfg(not(target_os = "windows"))]
        self.contexts.iter().for_each(|ctx| {
            if ctx.route_id != current_route_id {
                let pid = ctx.shell_pid;
                if pid > 0 {
                    teletypewriter::kill_pid(pid as i32);
                }
            }
        });

        #[cfg(target_os = "windows")]
        self.contexts.retain(|ctx| ctx.route_id == current_route_id);

        self.current_route = self.contexts[0].route_id;
        self.set_current(0);
    }
```

orhun:
Unfortunately that didn't work. The ttys are killed but the contexts are still there. 

See the gif:

rec_20240826T223312

raphamorim:
retain should have been enough

```
pub fn kill_pid(pid: i32) {
    unsafe {
        libc::kill(pid, libc::SIGHUP);
    }
}

impl Deref for Child {
    type Target = libc::c_int;
    fn deref(&self) -> &libc::c_int {
        &self.id
    }
}

impl Drop for Child {
    fn drop(&mut self) {
        unsafe {
            libc::kill(*self.pid, libc::SIGHUP);
        }
    }
}
```

drop isn't being called since the Context is under static, will refactor it to unblock this PR.

orhun:
works like a charm after rebasing on `main` 💯

raphamorim (approved):

raphamorim:
👌 🔥

Issue #335, opened 2023-11-11
Unable to get images working as expected


lypanov:
I've been trying to get the kitty/iTerm2 protocol support working and have tried timg (https://github.com/hzeller/timg) and imgcat script from iTerm2 itself but to no avail. I just see a single empty line instead of the image. Configuration: x11. Running from HEAD. Is there some TERM magic I'm missing maybe?

raphamorim:
Hey there @lypanov,

iTerm2 and kitty image protocol are coming to the next version only (I am still wrapping up the branch work before merge). ~Although the docs are showing there's image protocol available (my fault)~

lypanov:
Ah! That explains. Okay I'll await the branch then and continue using xfce4-terminal for his use case for now then switch that too in time.

Thanks for all the work on this. I'm running Rio via UserLand + Ubuntu on a Samsung Tab S6 and it works flawlessly for my needs.

raphamorim:
Next version will include an initial version of sixel and iterm2 protocol. Regarding kitty image protocol I will work in the future.

PR #625, opened 2024-08-30
Work in progress: Sixel protocol and iTerm2 image protocol

### LINKED ISSUE #38: Render images directly from the terminal
Back in time, terminals were always text-based.

Currently, almost everyone uses a terminal on a GUI environment. It amazes me to not having a terminal that doesn't render a image if I do a "cat sunset.jpg".

It would be nice if Rio could detect such output and renders it. Or maybe offering a built-in command to display them, maybe "draw image.jpg".

raphamorim:
Status: A lot of stuff to fix here yet...

demo

demo-sixel-2

closes #38 #335


raphamorim:
I will do iTerm2 image protocol in a separate PR, is easier for documentation later.

Kreijstal:
just compiled but it is not ported to windows yet, right?

PR #631, opened 2024-09-01
chore(config): rename option to hide-mouse-cursor-when-typing

### LINKED ISSUE #620: [config] hide-cursor-when-typing  option needs to be clearer that it is for the mouse cursor.
Hi

The hide-cursor-when-typing  option needs to be clearer that it's for the mouse cursor and not the terminal cursor.
May be rename to 

```
hide-mouse-cursor-when-typing = true
``` 

or change slightly to examples

```
mouse-cursor-typing = “hide”
mouse-cursor-typing = “show”

hide-mouse-cursor = “always”
hide-mouse-cursor = “never”
hide-mouse-cursor = “typing"

```

orhun:
closes #620

### DIFF
--- a/docs/docs/changelog.md
+++ b/docs/docs/changelog.md
@@ -15,6 +15,7 @@ language: 'en'
 - `use-kitty-keyboard-protocol` is now `true` as default.
 - Remove tokio runtime.
 - Allow configuring with lowercase values for enums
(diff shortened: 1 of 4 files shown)


raphamorim (approved):
👀 🔥

PR #632, opened 2024-09-01
refactor(config)!: revamp the cursor configuration

orhun:
> [!WARNING]  
> this is hella breaking

Before:

```toml
cursor = '▇'
blinking-cursor = false
```

After:

```toml
[cursor]
shape = 'block'
blinking = false
```


raphamorim:
I was going to write this code this week but you were quicker! 👀 

@orhun actually I think might be the best also change the style API to 

```toml
style = 'Block'
# or
style = 'Caret'
# or
style = 'underline'
```

What do you think?

orhun:
yup, makes sense. Will get to it soon 👍🏼

orhun:
I did you one better. Saw that there is already a `CursorShape` enum so I re-used it like so:

```toml
[cursor]
shape = 'block'
blinking = false
```

orhun:
Can you test it before merge? For me the other shapes were not working as expected (even without this PR) - will look into it.

raphamorim:
tested main, only `|` cursor working 👁️ 👁️ 

Screenshot 2024-09-05 at 19 39 19

orhun:
yeah for me too. Something is freaky.

Will check it out

raphamorim:
I figured out, I broke when I was refactoring SugarStyle into FragmentStyle lol
coming up a fix on main...

raphamorim:
Should be fixed now 

Screenshot 2024-09-05 at 20 02 56

rio-backend/src/ansi/mod.rs
```diff
@@ -27,6 +34,16 @@ impl CursorShape {
     }
 }
 
+impl From<CursorShape> for char {
+    fn from(value: CursorShape) -> Self {
+        match value {
+            CursorShape::Underline => '_',
+            CursorShape::Beam => '|',
+            _ => '▒',
```
raphamorim, line 42: I think the character is `'▇'`

raphamorim: actually we can get rid of this `From<CursorShape> for char`

raphamorim: and doesn't matter though which character, since it always fallbacks to block

orhun: 58b0c1b3094f1085e12a7827ab7e08495fd36aa3

raphamorim (approved):
🚀 ⭐

PR #649, opened 2024-09-06
feat(vi-mode): support search

### LINKED ISSUE #611: Support searching in Vim mode
It would be nice to be able use search in Vim mode via pressing `/`

Switching between Vim mode and search mode is not super convenient right now.

Any plans to support such feature?

orhun:
closes #611 

rec_20240906T173804

Checklist:

- [ ] Add key bindings for other platforms
- [ ] Check if the search highlight is working correctly
- [ ] Update the documentation

@raphamorim test it out and let me know :)


raphamorim:
I've tested locally and the triggers are correct but whenever I try to move to next or to back in the search then it leaves the search and return back vi mode only (example if I press enter).

orhun:
Yeah I experience the same issue. Committed ecd95b2ea482ccf11a6feb4566558564c127190d but it didn't fix...

I guess the search results are not highlighted too. Can you take a look and guide me a bit? 🙏

raphamorim:
sure! will take a look @orhun

raphamorim:
Merged into main 3815e7bd9564b45c26485274bda3ef2de2578811

PR #695, opened 2024-10-06
Delete the duplicated config key 'tabs-active-foreground'

bio:
Delete the duplicated config key 'tabs-active-foreground'.


raphamorim (approved):
Thank you!

PR #696, opened 2024-10-06
Fix hex size validation and add test cases for RGBA hex conversion

bio:
- Fixed regex that checks hex size.
- Added a test case to verify conversion of RGBA hex colors (for example, `#15151580`).

### DIFF
--- a/rio-backend/src/config/colors/mod.rs
+++ b/rio-backend/src/config/colors/mod.rs
@@ -735,7 +735,7 @@ mod tests {
     }
 
     #[test]
-    fn test_conversion_from_hex_with_alpha() {
+    fn test_conversion_from_gray_hex_with_alpha() {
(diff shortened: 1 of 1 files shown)


rio-backend/src/config/colors/mod.rs
```diff
@@ -738,4 +733,31 @@ mod tests {
             }
         );
     }
+
+    #[test]
```
raphamorim, line 737: That's great! thank you for the PR @bio , could add another test using `#06a49b99`?

bio: @raphamorim sure, I added an additional test case.

raphamorim: Thank you!

raphamorim (approved):

PR #710, opened 2024-10-16
Revert "windows: improve the performance for high polling rate mouse"

### LINKED ISSUE #708: [BUG] Window doesn't receive mouse events on Windows 11
The terminal's window does not receive any mouse events other than scrolling. So you can't resize the window, select text, etc. The only thing that works is clicking on the title bar to bring the window into focus.

It looks like the problem came from this commit: https://github.com/raphamorim/rio/commit/b4287fde7779206d43ffa685e49546920014bb11

From these lines, to be exact:
https://github.com/raphamorim/rio/blob/2231e255d82bbff8155594ba00f383070f292e7a/rio-window/src/platform_impl/windows/event_loop.rs#L2166-L2168

According to the [docs](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-nchittest), this branch must return values other than `HTNOWHERE` to work properly. I don't know much about the Win32 API, so I can't say how to do this correctly, but deleting this branch fixes all the problems with the mouse.

OS: Windows 11 (23H2)
OS Build: 22631.4317

igorsaux:
This reverts commit b4287fde7779206d43ffa685e49546920014bb11.

fix #708

https://github.com/user-attachments/assets/00fac33b-71b0-4018-8e2f-82b4de59cb9a


raphamorim (approved):
Thank you 👌

raphamorim (approved):

PR #713, opened 2024-10-16
fix: Set notepad as a default editor on windows

### LINKED ISSUE #677: Windows default editor shouldn't be "vi"
"vi" isn't installed by default on windows.

```toml
# Editor
#
# Default editor is "vi".
#
# Whenever the key binding `OpenConfigEditor` is triggered it will
# use the value of the editor along with the rio configuration path.
# [editor]
# program = "vi"
# args = []
```

This should probably be `notepad` as a default on Windows.

igorsaux:
That fixes the broken settings window on Windows. Creating a new blank settings window that then launches a separate notepad window looks weird, but I don't think there's a better solution.

fix #677

### DIFF
--- a/rio-backend/src/config/defaults.rs
+++ b/rio-backend/src/config/defaults.rs
@@ -84,9 +84,20 @@ pub fn default_theme() -> String {
 
 #[inline]
 pub fn default_editor() -> Shell {
-    Shell {
-        program: String::from("vi"),
-        args: vec![],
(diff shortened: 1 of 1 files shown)


raphamorim:
@igorsaux added this discussion in our discord that will be useful for future discussions on this change https://github.com/microsoft/terminal/discussions/16440

raphamorim (approved):

PR #722, opened 2024-10-19
gitignore: include relevant files

TornaxO7:
Those files are created from nix users. Would be nice, if `gitignore` would include them.


.gitignore
```diff
@@ -1,6 +1,10 @@
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
+# required by nix
+.direnv
+result
```
raphamorim, line 6: is result a folder? if yes, would you mind put as `result/`

TornaxO7: done

raphamorim (approved):

PR #731, opened 2024-10-26
Cursor blinking is by default off

melroy89:
Blinking is off by default, also according to the documentation: https://github.com/raphamorim/rio/commit/ee5bf32a43db5f10a6598fd963ac10f629217c43#diff-bb95aee1201743a824288d57d022dd0f24310b55146eeb475ec27fb3528b6f17R23

However, the toml config says it's on (true) by default. I think this is not correct.

---

Also fixed the family config indent issue.


raphamorim (approved):
Perfect, thank you

PR #746, opened 2024-10-30
Debian packages: ARM architecture

raphamorim:


.github/workflows/nightly.yml
```diff
@@ -74,21 +76,28 @@ jobs:
             LICENSE
 
   nightly-release-deb:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
andreban, line 79: Interesting - I'd assume cross-compilation would work on `ubuntu-latest`. Does build fail on it?

raphamorim: yea, i naively assumed ubuntu had support to arch but doesn't look it does yet

raphamorim: thinking about use something similar to this https://github.com/PMLS3/tauri-linux/blob/1992beef67640f2c843781bbb280f73116ad44c3/flows/arm.yml#L9

andreban: It seems the builds for linux are stalled on this PR. I'm thinking this may be the cause for it, given `ubunt-24.05-arm` doesn't seem to be one of GH's default runners - 
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

andreban: > thinking about use something similar to this https://github.com/PMLS3/tauri-linux/blob/1992beef67640f2c843781bbb280f73116ad44c3/flows/arm.yml#L9

Yeah, this should work. I'll try cross-compiling to ARM on an ubuntu box I have around and see what happens. Maybe it's just a matter of ensuring a few deps are installed.

raphamorim: Ah I see, I thought i got an error regarding cross compilation when installing cargo deb

raphamorim: @andreban if this is something you would like take to help i would glad let you take over 😂🫡

andreban: I didn't manage to build the project, but managed to figure some bits out. There's a dependency that needs the GCC compiler, so the compiler for aarch64 needs to be installed:

```
sudo apt install g++-aarch64-linux-gnu
```

Then
```
export RUSTFLAGS="-C linker=aarch64-linux-gnu-gcc"
cargo build --target aarch64-unknown-linux-gnu
```

But still fails

andreban: > @andreban if this is something you would like take to help i would glad let you take over 😂🫡

TBH, I likely won't have bandwidth to do it (work trip coming soon)

raphamorim: > I didn't manage to build the project, but managed to figure some bits out. There's a dependency that needs the GCC compiler, so the compiler for aarch64 needs to be installed:

I see, I will give another try after finish the split work. I think I got carried over 😆 (i thought i would finish this PR in 10-20min but it looks would need further investigation)

have a good time in your work trip <3

raphamorim: if you ever come to sweden lemme know as well ;)

.github/workflows/nightly.yml
```diff
@@ -74,21 +76,28 @@ jobs:
...
     permissions:
       contents: write
       discussions: write
 
+    strategy:
+      matrix:
+        target: [x86_64, aarch64]
+
     steps:
       - uses: actions/checkout@v3
```
melroy89, line 89:
```suggestion
      - uses: actions/checkout@v4
```

raphamorim:
Closing it in favour of #852

Issue #737, opened 2024-10-27
Failed to build from the latest git version on archlinux


HSidhiuwhfiunci:
I was using the PKGBUILD from the ABS with some modifications to make it possible to build from the latest git version, and here is the PKGBUILD and makepkg.conf:
[PKGBUILD.txt](https://github.com/user-attachments/files/17534252/PKGBUILD.txt)
[makepkg.conf.txt](https://github.com/user-attachments/files/17534253/makepkg.conf.txt)
It built successfully about one or two weeks ago, but always got failed during these several days.
And the failed log is like this:
[log.txt](https://github.com/user-attachments/files/17534289/log.txt)
(I am using the latest software from the archlinux which has the rust at 1.82.0 [1:1.82.0-2 for the actual rust pkg version from arch])

raphamorim:
Thank you for the issue!

interesting, i had same issue on macos whenever building release target. What fixed for me was use a slightly newer version of clang when compiling

orhun:
You need to build with LTO disabled:

```
options=('!lto')
```

raphamorim:
I am closing this regarding #752 

thank you 🙏

PR #748, opened 2024-11-02
Use as default thin lto instead to use full lto

NextWork123:
Hello, I noticed issue #737, and after some investigation, I would like to suggest switching from full LTO to ThinLTO. This change will optimize the compilation process more effectively than full LTO, resolve the compilation issues associated with it, and speed up the overall compilation time.

Best Regards,
NextWorks

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,3 +5,5 @@ rustflags = [
 ]
 runner = 'wasm-bindgen-test-runner'
 
+[target.x86_64-pc-windows-gnu]
+lto = true  
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,5 +5,12 @@ rustflags = [
 ]
 runner = 'wasm-bindgen-test-runner'
 
+[target.aarch64-apple-darwin]
+rustflags = [
+    # We can guarantee that this target will always run on a CPU with _at least_
+    # these capabilities, so let's optimize for them
+    "-Ctarget-cpu=apple-m1"
+]
+
 [target.x86_64-pc-windows-gnu]
 lto = true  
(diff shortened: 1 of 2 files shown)


orhun:
Any idea why thin LTO resolves the compilation issues on Arch Linux?

FWIW: https://github.com/briansmith/ring/issues/1444

NextWork123:
> Any idea why thin LTO resolves the compilation issues on Arch Linux?
> 
> FWIW: [briansmith/ring#1444](https://github.com/briansmith/ring/issues/1444)

some dependencies isn't passing with full lto (because of undefined symbols) so we need to switch to thin seem much packages on aur use thin as patch aswell i can put a patch in the pkgbuild and we can leave the full lto on all project errors about full lto:

https://paste.cachyos.org/p/eef0fd6.txt

let me know what i need to do

orhun:
Yup, I think this should be solved on the Arch side of things instead of upstream.

We can keep the fat LTO and use the workaround mentioned in <https://github.com/briansmith/ring/issues/1444> - that's what we do for most of the Arch packages these days.

NextWork123:
> Yup, I think this should be solved on the Arch side of things instead of upstream.
> 
> We can keep the fat LTO and use the workaround mentioned in [briansmith/ring#1444](https://github.com/briansmith/ring/issues/1444) - that's what we do for most of the Arch packages these days.

yeah, we can do that or switch the project to thin lto and remove the fat lto (isn't much safe as i said it do too lto on deps and other)

NextWork123:
I gonna close the pull request as resolve for atleast arch linux management we can leave fat lto as default and leave the package managers fix that.

PR #765, opened 2024-11-08
Add ability to specify platform specific shell.

### LINKED ISSUE #755: Multiplatform configuration
Currently Rio has no way to have different shell, window and navigation per OS.

So, we should introduce a new property called `platform`, that allow overwrite per OS.

This new property will allow users to configure on demand

```toml
[shell]
# default (in this case will be used only on MacOS)
program = "/bin/fish"
args = ["--login"]

[platform]
# Microsoft Windows overwrite
windows.shell.program = "pwsh"
windows.shell.args = ["-l"]

# Linux overwrite
linux.shell.program = "tmux"
linux.shell.args = ["new-session", "-c", "/var/www"]
```

crispy-strawberry:
Closes #755

### DIFF
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -899,7 +899,7 @@ impl<T: EventListener + Clone + std::marker::Send + 'static> ContextManager<T> {
         config: rio_backend::config::Config,
     ) {
         let (shell, working_dir) = process_open_url(
-            config.shell.to_owned(),
+            config.get_shell().to_owned(),
             config.working_dir.to_owned(),
             config.editor.to_owned(),
             None,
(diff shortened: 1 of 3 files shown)


melroy89:
Also add it to the example config file & documentation page?

raphamorim:
Thanks for the PR @crispy-strawberry !! I think there's just one thing, not only `Shell` should be under platform but also `Window` (https://raphamorim.io/rio/docs/config/window) and `Navigation` (https://raphamorim.io/rio/docs/config/navigation).

So probably the function you did `get_shell` might not be the best approach.

I would actually add a function on config called `overwrite_based_on_platform` here: https://github.com/raphamorim/rio/blob/main/frontends/rioterm/src/main.rs#L149-L181

This function would check if shell, window and navigation aren't none, if they are not then override each one.

What do you think?

raphamorim:
Added your changes and implemented the suggestion on cfbb148e 🙏

crispy-strawberry:
@raphamorim Sorry I'm a little busy with life atm and couldn't update it 😢

raphamorim:
No worrieeeeees at all!!! @crispy-strawberry <3

raphamorim:
I love the fact you started it, thank you for the contribution!

PR #774, opened 2024-11-12
docs: Add mention in homepage from mkenzo_8

marc2332:
image

```
Rio has the potential to be one of the fastest ⚡ and most customizable terminals out there, it is crazy.
```


raphamorim (approved):
<3

PR #779, opened 2024-11-13
Update window title straight away

hunger:
... instead of waiting for two seconds to do so.


hunger:
Unify the timeout handling for all platforms.

raphamorim (approved):

raphamorim:
Thank you!

Issue #801, opened 2024-11-19
Rio requires editor.args to be given in the config file, even if empty?


schneiderfelipe:
This crashes:

```toml
[editor]
program = "hx"
# args = []
```

This does not crash:

```toml
[editor]
program = "hx"
args = []
```

Version:

```console
❯ rio --version
rioterm 0.2.1
```

raphamorim:
No it shouldn't be a requirement, @Nylme made a fix for it 🙏 

thank you for the issue

PR #803, opened 2024-11-20
Makes editor.args and shell.args optional in config.toml

Nylme:
Fixes: https://github.com/raphamorim/rio/issues/801


raphamorim (approved):
Thank you, will merge on green 🙏

PR #807, opened 2024-11-20
Provide terminfo in the package on flake

baalajimaestro:
The flake currently is missing to pack the desktop file, and the terminfo.


raphamorim (approved):
looks good to me

PR #816, opened 2024-11-25
fix y-axis hint in config and docs

meowgorithm:
This little revision corrects the little helper text for the `padding-y` in the config. I was so confused!


raphamorim (approved):
⭐ thank you!!

PR #822, opened 2024-11-26
doc fixes (startup-directory → working-dir , SplitRigh → SplitRight)

raylu:
I also updated the `use-fork` link; not sure if it works, but it can't be more broken than the old one


raphamorim (approved):
perfect! thank you!!

Issue #832, opened 2024-12-04
Settings editor fails to open when `editor.program` is configured to use `micro` or `hx`


apcamargo:
When configuring the `editor.program` setting to use alternative editors [`micro`](https://github.com/zyedidia/micro) or [`hx`](https://github.com/helix-editor/helix), the settings editor fails to open via the keyboard shortcut. Attempts to add additional parameters have been unsuccessful.

### Steps to Reproduce
1. Change the `editor` setting to either `micro` or `hx` (after installing them).
2. Attempt to open settings using the keyboard shortcut (`⌘ + ,`).

### Observed Behavior
Settings editor does not open when using `micro` or `hx`.

### Expected Behavior
Settings editor should open with the configured editor.

### Additional Information
- `vim` and `vi` work properly.

raphamorim:
Hey @apcamargo thank you for the issue!!

Could be because you would need to place the full path of `hx`, could you try with the result of `where hx`?

apcamargo:
Good point

```sh
$which hx
/Users/apcamargo/.pixi/bin/hx
```

```toml
[editor]
program = "/Users/apcamargo/.pixi/bin/hx"
```

This worked. `micro` worked as well.

Is there a reason that `nano`, `vi`, and `vim` work without the full path? Maybe `micro` and `hx` would work as well if I put them in `/bin`?

raphamorim:
Yea, mostly is the reason. I think we can fix that by Rio side as well, but Rio would have to login in the shell (to acknowledge the path) then run the `$program`

I will close it for now but if you want to discuss more about it, we can convert this issue on discussion or ping me on Rio discord https://discord.gg/zRvJjmKGwS

apcamargo:
I think I'm ok as things are. It would be good to have this in the documentation somehow (or, better yet, built into Rio at some point)

raphamorim:
I agree 100%, is this something you would like to contribute? ( I think sometimes docs gets better written by people that had actually the issue itself )

apcamargo:
Yep. I can do that! I'm very busy the rest of the week, but I can work on
that right after.

raphamorim:
No worries, feel free to do whenever you have time! thank you 🙏

PR #841, opened 2024-12-08
Add warning about the editor path

apcamargo:
This PR addresses the discussion in https://github.com/raphamorim/rio/issues/832 by improving the documentation for the `editor` setting. It now mentions that the full path to the editor binary may be required in certain cases. Since I am unfamiliar with how Rio resolves binaries on Windows, the documentation changes are currently focused on Unix-like systems. Please feel free to expand or adjust it as needed!

As a side note, I think it's a good idea to keep that issue open if there are plans to revisit this behavior in the future. Many users likely expect that whatever is in their `PATH` should work seamlessly (at least, that’s my perspective).

### DIFF
--- a/docs/docs/config/editor.md
+++ b/docs/docs/config/editor.md
@@ -3,11 +3,11 @@ title: 'editor'
 language: 'en'
 ---
 
-Default editor is `vi`.
+This setting specifies the editor Rio will use to open the configuration file. By default, the editor is set to `vi`.
 
-Whenever the key binding `OpenConfigEditor` is triggered it will use the value of the editor along with the rio configuration path.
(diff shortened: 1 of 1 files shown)


raphamorim (approved):
Very good thank you

raphamorim:
> As a side note, I think it's a good idea to keep that issue open if there are plans to revisit this behavior in the future. Many users likely expect that whatever is in their PATH should work seamlessly (at least, that’s my perspective).

Let's see if it happens again after your change on docs, if does then I will do a better Rio warning on the terminal and include a message like "you filled with `nu` but maybe you mean `path/to/nu`"

PR #859, opened 2024-12-23
Redrawing window when moving around split screens

Ruivalim:
So, basically what this is doing is when you move around the split screens, it will trigger a re-render so only the cursor of the current split you are in will be filled and all the other will be hollow. Making it easier to know which split you currently are. 

Current behavior:
image

New behavior:
image

### DIFF
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -15,8 +15,9 @@ use renderable::RenderableContent;
 use rio_backend::config::Shell;
 use rio_backend::crosswords::{Crosswords, MIN_COLUMNS, MIN_LINES};
 use rio_backend::error::{RioError, RioErrorLevel, RioErrorType};
-use rio_backend::event::EventListener;
+use rio_backend::event::RioEventType;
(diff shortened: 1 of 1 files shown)


raphamorim:
Hey @Ruivalim thank you for the PR!

Isn't that working correctly? (At least it should be the default behaviour) Screenshot 2024-12-25 at 17 51 38

Lemme look the code changes

raphamorim:
aah ok, it's only for when you select next and prev using keybindings. I will think about the solution a bit more, I not a big fan of 2 sequenced render operations.

Btw, merry Christmas!

Ruivalim:
Hello @raphamorim, I didn't like the two render calls either, but my thought was that the two renders won't be a problem because, normally, you won't be switching tabs too often.

raphamorim:
Sorry for late reply on this, I had to rewrite a lot of stuff on screen past month due to performance fixes. This commit should address the issue 🙏 https://github.com/raphamorim/rio/commit/d8b9080e4004e5b419ec5c07c142cad26b7f3299

PR #884, opened 2025-01-08
docs: note possible log-level values

tzemanovic:
hi, thanks for rio! This is just a tiny PR to add a note with possible log-level values to the docs


raphamorim (approved):
Thank you!!

PR #886, opened 2025-01-09
Update windows-sys crate version to 0.59

andreban:
- Updates window.sys to version 0.58.
- A key change to this version is that the HANDLE type is now `*mut c_void`, so the code was changed to accomodate this.
- An important change is the manual implementation of the `Send` trait for `ChildExistWatcher`. The trait can no longer derived automatically because `*mut c_void` is not Send. However, pointers
are not Send more as a lint to developers (eg: a pointer to Rc<T>
would be safe), but `c_void` is Send and therefore it should be fine to derive it manually for the pointer.

See https://doc.rust-lang.org/nomicon/send-and-sync.html for details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3463,7 +3463,7 @@ dependencies = [
  "parking_lot",
  "signal-hook",
  "tracing",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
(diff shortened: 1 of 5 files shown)


raphamorim (approved):
Niiiicee! thank you 🙏

PR #898, opened 2025-01-12
feat: Sync text selection to available frames instead of 60ms

marc2332:
Am I missing something? 👁️ 👁️ 

Its wayyyyyyyyyyyyyyyyyyyy smoother now

Left: Latest nightly release
Right: This PR

https://github.com/user-attachments/assets/6d325deb-219a-45ec-aacc-518cb64197b2

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -668,7 +668,7 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     }
 
                     route.window.winit_window.set_cursor(CursorIcon::Pointer);
-                    route.window.screen.context_manager.schedule_render(60);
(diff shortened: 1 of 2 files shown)


raphamorim:
nice!
back in time it wasn't possible because request_render would trigger render like crazy, so that's why it was like that, now it has a control level under the scheduler.

will give it a test

raphamorim (commented):
Looks good, could you add details about it in the changelog (over docs folder)

https://raphamorim.io/rio/docs/releases/

raphamorim:
@marc2332 added the detail on main, thank you for the PR!

marc2332:
> @marc2332 added the detail on main, thank you for the PR!

cool!! 🦀 🔥

PR #901, opened 2025-01-14
fix(rioterm): inherit workspace readme for crate

hougesen:
Fixes the empty readme on crates.io 😄  

image


raphamorim (approved):
wow didn't know it was possible to do it

PR #906, opened 2025-01-15
fix: Relax libc requirements on ubuntu by compiling Rio in a 22.04 runner instead of 24.04

marc2332:
In theory, should avoid this:

imagen


.github/workflows/release.yml
marc2332, line 1: Lol, those extra spaces from below got removed on their own

raphamorim (approved):

PR #915, opened 2025-01-19
fix: Better handling of Escape and Enter keys

marc2332:
## Fixes
- No more need to press **twice** the Enter key in `ConfirmQuit`, `Welcome` or `Assistant` routes.
- Writing (and therefore pasting) was allowed while rendering the `Assistant` route. You could run stuff by pasting scripts with breaklines without the need to go to the terminal route

LMK if I broke something but from my own tests all went just fine


marc2332:
for context: the double enter issue was that the first press of Enter was not requesting a rerender, this was caused because `has_key_wait` was returnin `true` in all cases except when the route was Terminal, causing the first render not happen. The second was triggered simply because the route in that moment was the Terminal, and therefore an Enter in the shell causes a breakling, leading to a rerender

raphamorim (approved):
👌

PR #925, opened 2025-01-27
wip: Plugin System

raphamorim:


raphamorim:
Hey @nik-rev thanks for the comment! tbh is because i want to explore WASI and mostly because i have some good experience with WASM (used to contribute). Prob I am biased about wasm as well haha

PR #929, opened 2025-01-30
fix: Use `closesplitortab` instead of `closecurrenttaborsplit`

marc2332:


raphamorim (commented):
I am not sure if is actually a fix, but it's more like an improvement right?

Would be good to rename the action `CloseCurrentSplitOrTab` to `CloseSplitOrTab` as well

Also could document it in releases.md

marc2332:
> I am not sure if is actually a fix, but it's more like an improvement right?
> 
> Would be good to rename the action `CloseCurrentSplitOrTab` to `CloseSplitOrTab` as well
> 
> Also could document it in releases.md

Its actually a fix because the docs refer to it as `CloseSplitOrTab`

https://raphamorim.io/rio/docs/key-bindings#split-actions

raphamorim:
Ah then all good

raphamorim (approved):

PR #933, opened 2025-02-01
update default font style from normal to Normal

brunocroh:
## Update default font style on config and docs

Default font style values is writtend as 'normal' and 'italic' both throw an error because is expect that styles to be "Normal" and "Italic"

Screenshot 2025-02-01 at 10 04 56

### What I did

*Capitalize font style values in the docs and in `rio-backend/src/config/defaults.rs` 
  - Change "normal" to "Normal"
  - Change "italic" to "Italic"

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -210,25 +210,25 @@ features = []
 
 [fonts.regular]
 family = "cascadiacode"
-style = "normal"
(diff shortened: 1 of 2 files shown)


raphamorim (approved):
Thank you! ⭐

PR #936, opened 2025-02-02
Update broken link on generated config.toml

TeijiW:
I noticed that the link below is deprecated, as it results in a "not found" page:

https://github.com/raphamorim/rio/blob/9bfd12dff3dd69f72c6c3e7db0d48717259e9890/rio-backend/src/config/defaults.rs#L459

This PR is to update this link.


raphamorim (approved):
thanks!

PR #964, opened 2025-02-14
fix(goreleaser): special chars should be wrapped in `"`

vedantmgoyal9:


vedantmgoyal9:
my bad, i tested on main branch instead of the head branch :(

raphamorim (approved):

PR #982, opened 2025-02-22
ci: native macos notarization

caarlos0:
@raphamorim you'll need to set 2 more secrets:

```sh
gh secret set KEYCHAIN_PASSWORD --body "some password to lock and unlock the keychain"
gh secret set MACOS_NOTARY_KEYCHAIN_PROFILE --body "a profile name"
```

this uses the recently added native notarization: https://goreleaser.com/customization/notarize/#application-bundles-and-dmgs

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -119,10 +119,10 @@ notarize:
     - enabled: '{{ isEnvSet "MACOS_SIGN_P12" }}'
       sign:
         identity: "Developer ID Application: Hugo Amorim"
+        keychain: "{{ .Env.KEYCHAIN_PATH }}"
         options: [runtime]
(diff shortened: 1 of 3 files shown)


.goreleaser.yaml
```diff
@@ -249,32 +247,32 @@ winget:
 after:
   hooks:
     - cmd: brew bump-cask-pr rio --version {{ .Version }}
-      if: "{{ not .IsNightly }}"
+      if: "{{ .IsRelease }}"
```
caarlos0, line 250: this takes into account --snapshot as well - before these hooks would run if --snapshot was used

caarlos0:
cc/ @vedantmgoyal9 as well

PR #986, opened 2025-02-25
fix(goreleaser): bundle compiled terminfo in deb/rpm

vedantmgoyal9:


.goreleaser.yaml
```diff
@@ -151,10 +151,10 @@ nfpms:
     contents:
       - src: ./misc/rio.desktop
         dst: /usr/share/applications/rio.desktop
-      - src: ./misc/rio.terminfo
+      - src: ./misc/72/rio
```
devurandom, line 154: Why is it `72`?

vedantmgoyal9: i don't know honestly... but it remained same every time, i just hardcoded it. i **_think_** it will change when terminfo is modified, and i don't think it will happen anytime soon. i think @raphamorim can tell more about it

vedantmgoyal9: i don't know honestly... it remained same every time, so i didn't think too much about it and  just hardcoded it. i **_think_** it will only change when terminfo is modified, and i don't think it will happen anytime soon. maybe @raphamorim can tell more about it

raphamorim: `./misc/72/rio` is created after run `make install-terminfo`, that runs `@tic -xe rio -o $(APP_EXTRAS_DIR) $(TERMINFO)`.

I think if you run `tic` later in goreleaser prob is better use `./misc/rio.terminfo` instead

raphamorim (approved):

PR #995, opened 2025-02-28
Add Switch to Next/Prev Split or Tab command

vlabo:
This PR add two new commands
```rust
    /// Select next split if available if not next tab
    SelectNextSplitOrTab,
    /// Select previous split if available if not previous tab
    SelectPrevSplitOrTab,
```
This allows to switch between splits and tabs with single key binding.

I understand if you dont want to merge this and leave it for the plugin system, when available.
Thanks for making this great software.

### DIFF
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -1256,6 +1256,194 @@ pub mod test {
     }
 
     #[test]
+    fn test_switch_to_next_split_or_tab() {
+        let window_id: WindowId = WindowId::from(0);
+
+        let mut context_manager =
+            ContextManager::start_with_capacity(5, VoidListener {}, window_id).unwrap();
+        let should_redirect = true;
+        let split_down = false;
+
+        context_manager.add_context(should_redirect, 0);
+        context_manager.split(0, split_down);
+        context_manager.split(0, split_down);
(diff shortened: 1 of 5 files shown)


raphamorim:
Hey @vlabo thanks for the PR,

It does make sense! Could you update the docs folder regarding it: https://rioterm.com/docs/key-bindings#split-actions

frontends/rioterm/src/context/mod.rs
```diff
@@ -570,6 +570,32 @@ impl<T: EventListener + Clone + std::marker::Send + 'static> ContextManager<T> {
         self.current_route = self.current().route_id;
     }
 
+    #[inline]
+    pub fn switch_to_next_split_or_tab(&mut self) {
```
raphamorim, line 574: everything looks good, could you just add one test for next and prev so we make sure it always work 🙏

raphamorim:
Thank you for updating the docs @vlabo 🙏

frontends/rioterm/src/context/mod.rs
vlabo: Done. Thanks for the patience.

raphamorim (approved):
Nice! thank you @vlabo and thanks for the sponsorship too, I appreciate 🙏

PR #1003, opened 2025-03-05
Add missing LICENSE files to several workspace crates

musicinmybrain:
In the case of crates published on crates.io, like `rioterm`, `rio-proc-macros`, `teletypewriter`, etc., this fixes missing license text in the published crates. Normally, I would just use symbolic links to the top-level `LICENSE` file in a PR like this, but I followed the lead of existing license files like `corcovado/LICENSE` and made copies instead.

I also added the `LICENSE` file to `frontends/wasm`. It isn’t published on crates.io, and I don’t entirely understand how to use it, but I assumed it was the right thing to do there as well.

Including the license text with published crates is necessary so that people can comply with the chosen `MIT` license, which says:

https://github.com/raphamorim/rio/blob/73f0158330a0fe6d639406fe748588dc1bb69816/LICENSE#L12-L13

### DIFF
--- a/frontends/rioterm/LICENSE
+++ b/frontends/rioterm/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022-present Raphael Amorim
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


raphamorim (approved):
Thank you @musicinmybrain

PR #1012, opened 2025-03-14
Document git-lfs being required to build from source (missing assets)

laundmo:
While trying to build from source for the first time, i ran into needing to install git-lfs as otherwise some font files were invalid. This just adds a note to docs regarding this.

### DIFF
--- a/docs/docs/install/build-from-source.md
+++ b/docs/docs/install/build-from-source.md
@@ -10,6 +10,8 @@ Before compiling Rio terminal, you'll have to first clone the source code:
 ```sh
 git clone https://github.com/raphamorim/rio.git
 ```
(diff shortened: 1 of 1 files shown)


raphamorim:
Hmmm is true, I forgot by tracking font files lfs would become a dependency. idk how I about it, the website it does make sense (due to large files) but the core code i don't think so.

Could you actually remove fonts from lfs?

laundmo:
Yeah, removing them is probably possible. Though i wonder if it would make sense to reduce the amount of fonts required to build, i don't know what they're all used for.

raphamorim:
Those are fallback fonts, originally Rio didn't have any but let's just say having those makes my life as maintainer easier 😆

laundmo:
Hmm, does it need that many? since they're compiled in, wouldn't one be enough?

PR #1025, opened 2025-03-30
Add flake install instructions

datsfilipe:
# Description

Not sure if it's a wanted change, but I found interesting to add since I was able to install it using the repo flake but did not found any installation instructions for it.

# Changes in this PR

- Removed duplicated `cursor` item in `colors` config example. Got me a configuration error to notice.
- Added flake installation instructions for NixOS modules, home-manager modules and installation to a profile.


raphamorim:
Very cool! thanks for the PR!

I am not a nix user myself, but will ask @TornaxO7 and @RaySlash to review your PR instead 🙏

RaySlash:
Hey, This looks pretty good to me. Flake install instructions were not really much used and most of the new user should use the nixpkgs derivation which is going to be stable for them than pinning to flake. Instructions never hurt and could you add some lines along the lines of "If you are unsure to use the nixpkgs or flakes package, always go for the nixpkgs derivation as the flakes package output is for development purposes only". 😸

datsfilipe:
Added a note among these terms. @RaySlash

RaySlash (approved):

docs/docs/config.md
```diff
@@ -43,7 +43,6 @@ Example:
 background = '#0F0D0E'
 black = '#4C4345'
 blue = '#006EE6'
-cursor = '#F38BA3'
```
raphamorim, line 46: Any reason to remove cursor?

raphamorim: Ah it’s duplicated!

raphamorim (approved):
Thank you!!

PR #1030, opened 2025-04-02
README: make repology badge display 3 columns

luzpaz:
Preview: https://github.com/luzpaz/rio/blob/repology/README.md#packaging


raphamorim (approved):
Thank you, looks better

PR #1074, opened 2025-04-21
Remove duplicate cursor in default-colors.md

theJian:
Remove duplicate key `cursor` in table `colors`, which throws when pasted into the configuration.


raphamorim (approved):
thank you!

PR #1082, opened 2025-04-25
Add support for $RIO_CONFIG_HOME and additional font dirs

TheSylex:


sugarloaf/src/font/fonts.rs
```diff
@@ -141,6 +141,8 @@ pub struct SugarloafFonts {
     pub symbol_map: Option<Vec<SymbolMap>>,
     #[serde(default = "bool::default", rename = "disable-warnings-not-found")]
     pub disable_warnings_not_found: bool,
+    #[serde(default = "Option::default")]
```
raphamorim, line 144: please add `rename = "additional-dirs"`

sugarloaf/src/font/mod.rs
```diff
@@ -275,6 +275,21 @@ impl FontLibraryData {
         let mut db = loader::Database::new();
         db.load_system_fonts();
 
+        let additional_dirs = spec
+            .additional_dirs
+            .unwrap_or_default()
+            .into_iter()
+            .map(PathBuf::from);
+
+        let config_home_dir = std::env::var("RIO_CONFIG_HOME")
```
raphamorim, line 284: I would avoid add this functionality now, I need to think more about this solution tbh. And prob will be good to release Rio with `RIO_CONFIG_HOME` first.

There's few concerns I have so far regarding env vars usage but most importantly is that sugarloaf shouldn't be connected to rio whatsoever since is used by other projects.

raphamorim:
Nice!! left few comments, overral looks good!

Could you also add information about it in https://rioterm.com/docs/releases ?

raphamorim (approved):

PR #1099, opened 2025-05-09
Normalize font metrics, scaling the rendering height of all characters to the same spacing.

### LINKED ISSUE #1071: CJK characters display "higher" than Latins, so the terminal scroll to wrong place after showing long CJK text
CJK characters in rio seem to displays "higher" than Latins. It's a good idea, because CJK characters become more readable with more vertical spaces. (In contrast, that in vim is less readable)

Image

However, I found that after a long CJK text was output in rio terminal, it run into a strange status...

Image

If all characters are Latin, everything works well like the top-left terminal shows.

But when characters are all CJK or CJK/Latin mixed, as the right terminal shows, the terminal scrolls **less than 4 lines down** after the `printf` command executed.

I think the problem is caused by rio assuming the height of contents is equal to that of **4 Latin character lines**, not **4 CJK/Latin mixed character lines**.

Image

Here are two examples with two passages, one is [*Li Sao*](https://zh.wikisource.org/zh-hans/%E9%9B%A2%E9%A8%B7) in Chinese, the other is a part of [*Hamlet*](https://en.wikisource.org/wiki/Hamlet_(1917)_Yale/Text/Act_III). After `cat Hamlet` everything goes well, but after `cat "Li Sao"` it scrolls to quite a wrong place and I cannot even see the input line!

Tryanks:
Fix #1071

### DIFF
--- a/sugarloaf/src/font_introspector/metrics.rs
+++ b/sugarloaf/src/font_introspector/metrics.rs
@@ -178,6 +178,24 @@ impl MetricsProxy {
                 self.leading = hhea.line_gap();
             }
         }
+        const DEFAULT_SPACING: i16 = 2380;
+        let current_sum = (self.ascent + self.descent + self.leading) as f32;
+        if current_sum != 0.0 {
+            let scale = DEFAULT_SPACING as f32 / current_sum;
+            self.ascent = (self.ascent as f32 * scale) as i16;
+            self.descent = (self.descent as f32 * scale) as i16;
+            self.leading = (self.leading as f32 * scale) as i16;
+
+            let final_sum = self.ascent + self.descent + self.leading;
+            if final_sum != DEFAULT_SPACING {
+                let diff = DEFAULT_SPACING - final_sum;
+                if diff > 0 {
+                    self.leading += diff;
+                } else {
+                    self.descent += diff;
+                }
+            }
+        }
         let vhea = font.vhea();
         if let Some(vhea) = vhea {
             self.has_vertical_metrics = true;


sugarloaf/src/font_introspector/metrics.rs
```diff
@@ -178,6 +178,24 @@ impl MetricsProxy {
                 self.leading = hhea.line_gap();
             }
         }
+        const DEFAULT_SPACING: i16 = 2380;
```
raphamorim, line 181: sorry for take long to look, i was a bit sick 🤧 
where is that value coming from?

Tryanks: This actual value comes from my real-world measurement: On my macOS, I output `self.ascent`, `self.descent`, and `self.leading`, and obtained 1900, 480, and 0 respectively for Latin characters. Adding them together gives this absolute value.

_I closed this PR just due to an unexpected mistake: I wanted to change the branch name "main" in the merge source to an actual description, but I didn't expect this to trigger it. If it can be accepted, I'm willing to open a new PR anytime._

Tryanks: If each font is measured individually, it would bring us back to the starting point—the current code assuming this PR doesn't exist.

I believe that for a terminal emulator, the line height and baseline position of each line in the rendered display should be fixed, not dynamically calculated based on the font.

When processing each rune, aligning baselines across different fonts and scaling ascenders/descenders to fit the line height may be the ideal rendering approach.

raphamorim: The problem is that fixed numbers can break for different type of fonts, like fira code and cascadia mono have different values for the same height. Due to char spacing as well.

Ideally the best would be fix the calculation for retrieve that number

PR #1100, opened 2025-05-09
Update build-from-source.md

wellyton-xs:
Add --depth=1 in "git clone" to avoid dealing with a very long download time that can annoy beginners and new users.


raphamorim (approved):
Thanks!

PR #1105, opened 2025-05-13
Read configuration from $XDG_CONFIG_HOME on Linux

ranisalt:
Uses `$XDG_CONFIG_HOME` to find configuration files on Linux, with a lower priority than `$RIO_CONFIG_HOME` but higher than `$HOME` as specified in [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/)

> There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable `$XDG_CONFIG_HOME`.

The fallback to `$HOME/.config` is specified further below:

> `$XDG_CONFIG_HOME` defines the base directory relative to which user-specific configuration files should be stored. If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.

Previous work at #14 and also needed for flathub/flathub#6483

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -5,7 +5,9 @@ language: 'en'
 
 The configuration should be the following paths otherwise Rio will use the default configuration.
 
-MacOS and Linux configuration file path is `~/.config/rio/config.toml`.
+MacOS configuration file path is `~/.config/rio/config.toml`.
+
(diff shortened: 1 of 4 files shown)


raphamorim (approved):

raphamorim:
@ranisalt could you update the releases.md file in the docs folder

raphamorim (approved):
⭐

PR #1106, opened 2025-05-14
Add metainfo and screenshots for appstream

ranisalt:
Metainfo is a common format for metadata used by application stores, such as Flathub and GNOME Software, to display information about an app. The policy is that metainfo should reside and be maintained in the application repository, not by the store, as requested on flathub/flathub#6483

I added the changelogs for the past 3 published versions (which is optional and can be removed, but nice to have) and the history of the versions > 0.1.0 to keep track of the publishing dates.

Opening as draft until I get a green light on the quality of the screenshots to avoid having to replace them in the future.


docs/static/assets/appstream/default.png
ranisalt, line 1: Should these files be added to LFS? I suppose they won't change too much

PR #1108, opened 2025-05-15
Better lock that cargo install

EsfoNL:
without `--locked` its borken because cargo install doesnt respect the lock file without it


raphamorim (approved):
⭐

PR #1109, opened 2025-05-15
Adding Shader Parameter Config to Rio

JRBros2346:


JRBros2346:
Updated `librashader` from `0.7.1` to [`passmeta-preset`](https://github.com/SnowflakePowered/librashader/tree/passmeta-presets) branch
Also replaced `sugarloaf\src\components\filters\runtime` with `librashader-runtime-wgpu`. Latest version supports float16, so it should not be a major problem. 
Due to the previous change, i had to remove [this line](https://github.com/JRBros2346/rio/blob/10403f7e23ac3f2a7b2059dd1c177919400e8344/sugarloaf/src/components/filters/mod.rs#L237) I don't know its significance, but hopefully its not a major problem.
Huge thanks to @chyyran for helping me out through this process.

Enumerating all the Shader Parameters now works AWESOME ✨

sugarloaf/src/components/filters/mod.rs
```diff
@@ -197,24 +218,23 @@ impl FiltersBrush {
...
+                ctx.format,
+            );
             let dst_viewport =
                 Viewport::new_render_target_sized_origin(dst_output_view, None).unwrap();
 
             // Framecount should be added forever: https://github.com/raphamorim/rio/issues/753
             self.framecount = self.framecount.wrapping_add(1);
             if let Err(err) = filter.frame(
-                filter_src_texture,
+                &*filter_src_texture,
```
raphamorim, line 232: was this a change in the library?

JRBros2346: Yes, thats right. `FilterChainWgpu::frame` requires a `&texture` instead of an `Arc<Texture>`

JRBros2346:
Since librashader doesn't require an `Arc` reference anymore, i have removed those and replaced them with `&` references.

PR #1116, opened 2025-05-19
Skip passing sandbox env to spawned shell in Flatpak

ranisalt:
When passing `std::env::vars_os()` to the spawned shell inside a Flatpak sandbox, it ends up overriding important host environment variables with overridden values meant for the sandboxed application (in this case, Rio). Most notably, `$XDG_CONFIG_HOME` is overridden by the value passed to Rio, which is `$HOME/.var/app/com.rioterm.Rio/config` instead of `$HOME/.config` (or whatever value the user has set):

image

(current behavior)

There's no need to pass any environment variables to the shell outside the sandbox, they get inherited from the process that spawned the Flatpak sandbox, which is usually the user session that already has the environment properly set up:

image

(fixed behavior)

Additionally, the order of the parameters does not matter in this scenario, so I moved setting `TERM_PROGRAM` (now `TERM` and `COLORTERM`, since `TERM_PROGRAM` is for Mac) to the initial data of the vector and skip the push.

As a bonus, it now renders icons and ligatures as the shell has access to the host fonts :)

image image

### DIFF
--- a/docs/docs/releases.md
+++ b/docs/docs/releases.md
@@ -7,7 +7,7 @@ language: 'en'
 
 ## 0.2.17 (unreleased)
 
-- TBD.
+- Skip passing sandbox env in Flatpak, fixes user environment in spawned shell [#1116](https://github.com/raphamorim/rio/pull/1116) by [@ranisalt](https://github.com/ranisalt).
 
 ## 0.2.16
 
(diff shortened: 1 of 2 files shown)


raphamorim (approved):
:ok

raphamorim:
@ranisalt nice! make sense, could you also update the releases.md

PR #1133, opened 2025-06-06
fixed documentation for default values of window.decorations fixes #1132 

### LINKED ISSUE #1132: default value of window.decoration for macos does not match documentation
According to the documentation at: https://rioterm.com/docs/config#windowdecorations the default value for `window.decoration` is `Transparent` for macOS. However, this does not seem to be true for `v0.2.17` and beyond. When I upgraded to `v0.2.17` the top bar became gray instead of matching the terminal background. I had to manually set `window.decoration = "Transparent"` on my macos. 

Not sure if others can reproduce this behavior on macOS but I thought i should document this.

ajinkya-k:
The default value of `window.decorations` on macOS was changed to enabled in `v0.2.17`. This is documented in the release notes but was not changed in the online documentation. This commit fixes that discrepancy.

See release notes too: 
https://github.com/raphamorim/rio/blob/7ef143fd3a5de979a912e181882fc1de97be6460/docs/docs/releases.md?plain=1#L18

fixes #1132

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -1454,9 +1454,9 @@ y = 0.0
 
 Set window decorations.
 
-- `Enabled` (default for Windows/Linux/BSD) enable window decorations.
+- `Enabled` (default for Windows/Linux/BSD/macOS) enable window decorations.
 - `Disabled` disable all window decorations.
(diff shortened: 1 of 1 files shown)


raphamorim:
Yes, thank you very much @ajinkya-k !!

Issue #1115, opened 2025-05-18
Underline height makes text hard to read


nik-rev:
When text is underlined, the underline interferes with the text making it hard to read

Example:

![image-20.png](https://github.com/user-attachments/assets/a463ceee-ea19-407d-ae2b-8e496e882fe8)

This is coming from my editors diagnostics (helix).

This is WezTerm:

![image-28.png](https://github.com/user-attachments/assets/e2db2050-298b-4eed-b936-52f96a649c1d)

As you can see the difference in text readability is quite noticeable

Would be nice to have some space between the underline and the text so its more readable

(and potentially making this configurable)

PR #1141, opened 2025-06-10
fix: underline height making text hard to read

nik-rev:
Currently, underline height is too high which makes text harder to read. The PR fixes that by rendering them 5px lower.

Before | After

image

This example shows an invalid snippet of code rendering the underlines

WezTerm (for reference) | Rio Before this PR | Rio With this PR 

image

Closes https://github.com/raphamorim/rio/issues/1115 

---

Note: The PR isn't really ready for merge
- I think decreasing it further by 1px or 2px will yield better results
- At the moment the underlines get clipped, as you can see their bottom part does not get rendered. I'm trying to figure out how to let them be drawn out of bounds of the cells

### DIFF
--- a/frontends/rioterm/src/renderer/mod.rs
+++ b/frontends/rioterm/src/renderer/mod.rs
@@ -29,6 +29,9 @@ use std::ops::RangeInclusive;
 use rustc_hash::FxHashMap;
 use unicode_width::UnicodeWidthChar;
 
+/// Draw the underline this amount of pixels below the text
+pub const UNDERLINE_OFFSET: f32 = -6.0;
+
 #[derive(Default)]
 pub struct Search {
     rich_text_id: Option<usize>,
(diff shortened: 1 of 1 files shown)


nik-rev:
I think to fix the clipping issue I'll need to adjust something here:

```rs
UnderlineShape::Curly => {
    let style_line_height = (line_height / 10.).clamp(2.0, 16.0);
    let size = (style_line_height / 1.5).clamp(1.0, 4.0);
    let offset = style_line_height * 1.6;

    let mut curly_width = ux;
    let mut rect_width = 1.0f32.min(end - curly_width);

    while curly_width < end {
        rect_width = rect_width.min(end - curly_width);

        let dot_bottom_offset = match curly_width as u32 % 8 {
            3..=5 => offset + style_line_height,
            2 | 6 => offset + 2.0 * style_line_height / 3.0,
            1 | 7 => offset + 1.0 * style_line_height / 3.0,
            _ => offset,
        };

        self.add_rect(
            &Rect::new(
                curly_width,
                uy - (dot_bottom_offset - offset),
                rect_width,
                size,
            ),
            depth,
            &underline.color,
        );

        curly_width += rect_width;
    }
}
```

`sugarloaf/src/components/rich_text/batch.rs:3880` in function `BatchManager::draw_underline`

raphamorim:
Thank you for your PR it was supposed to fixed in the latest 🙏

PR #1203, opened 2025-07-07
Use the first font (not the first rune) as the rendering reference.

### LINKED ISSUE #1071: CJK characters display "higher" than Latins, so the terminal scroll to wrong place after showing long CJK text

Tryanks:
Fix #1071 .
截屏2025-07-07 18 24 32

### DIFF
--- a/sugarloaf/src/components/rich_text/mod.rs
+++ b/sugarloaf/src/components/rich_text/mod.rs
@@ -375,8 +375,15 @@ impl RichTextBrush {
 
     fn extract_font_metrics(
         lines: &[crate::layout::BuilderLine],
+        font_library: &FontLibrary,
+        primary_font_size: f32,
     ) -> Option<(f32, f32, f32, usize, f32)> {
-        // Extract the first run from a line that has at least one run
+        // Try to get metrics from primary font first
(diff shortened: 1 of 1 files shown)


Tryanks:
Has been fixed in #1208

PR #1213, opened 2025-07-13
fix: #1139  vi cursor not displayed when moving

### LINKED ISSUE #1139: VI cursor not visible
Cursor does not show when in VI mode. Cursor is visible only when in selection mode (pressing V). Selection and yank properly works.
Running Rio + Nushell + Starship, but same happens using bash.

https://github.com/user-attachments/assets/2d6b74f3-dd55-41cc-88b6-9a3801191a86

Malandril:
Fixes #1139 , the lines the vi cursor was moving to were not to be updated.


raphamorim:
Hey @Malandril thanks for the PR! Will take a look this week 🙏

Malandril:
The modified code, is not used any more on main, see #1380

PR #1217, opened 2025-07-14
canonicalize the input to the --working-dir argument for rio cli

### LINKED ISSUE #1210: rio --working-dir commandline option does not work correctly with relative paths
Hello! 

Sometimes I invoke a rio terminal form inside vs code terminal. I tried using:
```bash
rio --working-dir=.
```
but that opens rio in the user home directory, however
```bash
rio --working-dir=$(PWD)
```
works correctly. 

I tried 
```bash
rio --working-dir=..
```
and it opens in `/Users` thats when i figured out that for some reason it treats user home as the `.` instead of the actual current working directory. Is this expected behavior? Am I missing something? 

OS: macOS 15.5
arch: aarch64 (Apple Silicon M3Pro)

ajinkya-k:
When rio is called from a different terminal as:
```
rio --working-dir=<some path>
```
providing a relative path doesn't work correctly and in fact the relative path is treated as relative from the home directory 
See the issue #1210.

This PR canonicalizes the input path before the new rio terminal is created.

fixes #1210

### DIFF
--- a/frontends/rioterm/src/main.rs
+++ b/frontends/rioterm/src/main.rs
@@ -179,21 +179,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
 
         if let Some(working_dir_cli) = args.window_options.terminal_options.working_dir {
-            config.working_dir = match std::fs::canonicalize(&working_dir_cli).and_then(|path| {
-                if path.is_dir() {
-                    path.into_os_string().into_string().map_err(|_| {
-                        std::io::Error::new(
-                            std::io::ErrorKind::InvalidData,
-                            "Invalid UTF-8 in path",
-                        )
-                    })
-                } else {
-                    Err(std::io::Error::new(
-                        std::io::ErrorKind::NotADirectory,
-                        "Path is not a directory",
-                    ))
-                }
-            }) {
+            config.working_dir = match std::fs::canonicalize(&working_dir_cli).and_then(
+                |path| {
+                    if path.is_dir() {
+                        path.into_os_string().into_string().map_err(|_| {
+                            std::io::Error::new(
+                                std::io::ErrorKind::InvalidData,
+                                "Invalid UTF-8 in path",
+                            )
+                        })
+                    } else {
+                        Err(std::io::Error::new(
+                            std::io::ErrorKind::NotADirectory,
+                            "Path is not a directory",
+                        ))
+                    }
+                },
+            ) {
                 Ok(canonical_path) => Some(canonical_path),
                 Err(e) => {
                     tracing::warn!("Failed to set working directory '{}': {}. Using default instead.", working_dir_cli, e);


ajinkya-k:
@raphamorim i tested it by trying a few relative paths and it works correctly. I dont know how else to test it. Any ideas?

ajinkya-k:
> could you also update the releases.md

certainly!

ajinkya-k:
would this be considered a breaking change?

frontends/rioterm/src/main.rs
```diff
@@ -179,7 +179,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
 
         if let Some(working_dir_cli) = args.window_options.terminal_options.working_dir {
-            config.working_dir = Some(working_dir_cli);
+            let wd_canonical = match std::fs::canonicalize(&working_dir_cli)?
```
raphamorim, line 182: prob better:

```rust
let wd_canonical = match std::fs::canonicalize(&working_dir_cli)
    .and_then(|path| {
        if path.is_dir() {
            path.into_os_string().into_string().map_err(|_| 
                std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid UTF-8 in path"))
        } else {
            Err(std::io::Error::new(std::io::ErrorKind::NotADirectory, "Path is not a directory"))
        }
    })
{
    Ok(canonical_path) => Some(canonical_path),
    Err(e) => {
        tracing::warn!("Failed to set working directory '{}': {}. Using default instead.", working_dir_cli, e);
        None
    }
};
```

frontends/rioterm/src/main.rs
```diff
@@ -179,7 +179,30 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                        })
+                    } else {
+                        Err(std::io::Error::new(
+                            std::io::ErrorKind::NotADirectory,
+                            "Path is not a directory",
+                        ))
+                    }
+                },
+            ) {
+                Ok(canonical_path) => Some(canonical_path),
```
raphamorim, line 199: one more suggestion, let's get rid of the variable, something like:

match std::fs::canonicalize(&working_dir_cli)
    .and_then(|path| {
        if path.is_dir() {
            path.into_os_string().into_string().map_err(|_| 
                std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid UTF-8 in path"))
        } else {
            Err(std::io::Error::new(std::io::ErrorKind::NotADirectory, "Path is not a directory"))
        }
    })
{
    Ok(canonical_path) => {
        config.current_working_directory = canonical_path;
    },
    Err(e) => {
        tracing::warn!("Failed to set working directory '{}': {}. Using default instead.", working_dir_cli, e);
        None
    }
};

ajinkya-k:
The nix build failed and I don’t know anything about nix 😢

raphamorim:
No worries, it has been failing since I updated Rust version. Thank you !

PR #1233, opened 2025-07-23
gitignore: fixes for nix output links

spikespaz:
- `result/` is a directory, which was supposed to be a symbolic link (file), without trailing `/`.
- `result-lib`, `result-dev` and `result-man` are valid Nix build outputs that should also be ignored.
- Some files should only realistically appear in the root of the repository, for those names, prefix with `/` to prevent future confusion.

There's other junk in the `.gitignore` that looks like it started from a (very bad) template. I'm not sure what else should be removed, but I see very few projects ignore `rustmft` backup files (it is stable enough that these should almost never be created) and `rls` is deprecated. I haven't touched these entries that I think should be removed outright.


spikespaz:
@raphamorim I have more to contribute, but I'm wondering if these changes are welcome at all.

spikespaz:
Separate changes at the same time. I noticed the gitignore was messed up when I was testing my changes.

spikespaz:
My original message has a poorly phrased question however, do you want me to add another commit removing whatever I thought was useless?

.gitignore
```diff
@@ -1,11 +1,13 @@
...
-# required by nix
-.direnv/
-result/
+# Nix build output
+/result
+/result-lib
+/result-dev
+/result-man
 
-target/
```
raphamorim, line 8: This one I rather keep because I use rio in a private fork that I have been working and target is independent there for multi levels

spikespaz: Why would that be the case? The repository is a workspace, and thus, `target` should only appear at `/`.

Having a directory/module named `target` is perfectly reasonable, perhaps a compromise would be to add `!/src/**/target`?

Or, I'm a big fan of adding `.gitignore` in directories which need special handling (for offline work, for example).

.gitignore
```diff
@@ -1,11 +1,13 @@
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
-# required by nix
-.direnv/
-result/
+# Nix build output
+/result
```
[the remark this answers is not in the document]
spikespaz: This `/` prefix excludes matches only at the root of the repo. There is no trailing `/` because it's always a symlink, so a directory named `result` is fine.

You could use `result` as a bare name, but I don't like to make `.gitignore` patterns unnecessarily strict.

PR #1234, opened 2025-07-23
flake: devShells+packages: fix rust-overlay toolchain usage

spikespaz:
You don't need a custom `rust-toolchain` if you aren't actually adding any extensions that aren't already present.

- For the `devShells.*.msrv` output, it was never using the project's `rust-toolchain.toml` (that is, assuming you use it for the MSRV toolchain). Now it is.
- For the `packages.*.rio` output, you can just use `rust-bin.stable.latest.minimal`. That's all that is required to build with `rustPlatform.buildRustPackage`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -56,7 +56,7 @@
         };
         packages.rio = pkgs.callPackage mkRio {rust-toolchain = rust-toolchain;};
 
-        devShells.msrv = mkDevShell rust-toolchain;
+        devShells.msrv = mkDevShell (pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml);
         devShells.stable = mkDevShell pkgs.rust-bin.stable.latest.default;
         devShells.nightly = mkDevShell (pkgs.rust-bin.selectLatestNightlyWith (toolchain: toolchain.default));
       };
--- a/flake.nix
+++ b/flake.nix
@@ -26,10 +26,6 @@
         lib,
         ...
       }: let
-        rust-toolchain = pkgs.rust-bin.stable.latest.default.override {
-          extensions = ["rust-src" "rust-analyzer"];
-        };
-
         mkRio = import ./pkgRio.nix;
 
         mkDevShell = rust-toolchain: let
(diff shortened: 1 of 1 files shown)


spikespaz:
@TornaxO7 Requesting your review.

spikespaz:
By the way, @raphamorim, I would really appreciate if you rebase-merged my commits instead of squashing or creating a merge commit. I take care to keep organized commits, they're ready to be committed to `main` directly.

flake.nix
```diff
@@ -54,9 +50,9 @@
           type = "app";
           program = self'.packages.default;
         };
-        packages.rio = pkgs.callPackage mkRio {rust-toolchain = rust-toolchain;};
+        packages.rio = pkgs.callPackage mkRio {rust-toolchain = pkgs.rust-bin.stable.latest.minimal;};
```
TornaxO7, line 53: Fair enough, I can see the reasoning behind this one.
:+1:

flake.nix
```diff
@@ -54,9 +50,9 @@
...
 
-        devShells.msrv = mkDevShell rust-toolchain;
+        devShells.msrv = mkDevShell (pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml);
```
TornaxO7, line 55: Hm... we can do this but then I'd like to add the `rust-analyzer` component to the toolchain (`rust-toolchain.toml`). What do you say?

TornaxO7 (commented):
Sorry for the delay. Here's my review.

flake.nix
TornaxO7: > I disagree. If rust-toolchain.toml represents the minimal set of Rust software required to build the project, rust-analyzer is not required.

True. Sounds reasonsble. According to this SO answer (https://stackoverflow.com/a/62419829) it does look like as if the main purpose of the rust toolchain is to install the _required_ tools to build the rust project and that you can optionally add other compononents.

So yeah, adding `rust-analyzer` and `rustfmt` just to the devshell if fine for me.

I'd like to try your PR later out and then my approvement if everything works 👍

TornaxO7 (approved):

raphamorim (approved):
Thank you!

PR #1251, opened 2025-08-04
Add install instructions for scoop

IceSentry:
I personally prefer using scoop over chocolatey or winget and the package is already available but it's not mentioned in the install instructions for windows

Side note, chocolatey is mentioned but it only has a _very_ outdated version of rio from december 2023. The scoop version was updated yesterday so it's way more up to date.


raphamorim:
Make sense! Thank you 🙏

PR #1281, opened 2025-08-21
Store fonts after loading them to prevent aggressive re-loading

itsjunetime:
Right now on main, trying to quickly print a lot of emojis while using an emoji-heavy font introduces a massive amount of lag. This can be easily demonstrated if one sets up neovim to use [folke/drop](https://github.com/folke/drop.nvim) (which "drops" a lot of emojis from the top of the screen as a screensaver) and runs it while using a custom emoji font.

This lag comes from the fact that `FontLibraryData::get_data` doesn't store the data of loaded fonts after it loads it. This means that, if I'm understanding correctly, this function is essentially thrashed each time one of these emojis is loaded, as the whole font has to be loaded from disk time and time again.

So this PR simply stores the font data into `FontLibraryData` after loading it. I tested it and it makes the lag completely go away and significantly reduces resource utilization.

This also makes some small nit changes that I noticed while looking for this fix, such as the `let _ else` and changing `SharedData` to just hold a `Arc<[u8]>` since that's all it actually ever holds.

I don't know if this is the best way to introduce this fix, but it seems to do the job just fine for me. I'm totally happy to close this in favor of a different PR if there's a better way to achieve the same results.

### DIFF
--- a/frontends/rioterm/src/renderer/font_cache.rs
+++ b/frontends/rioterm/src/renderer/font_cache.rs
@@ -113,7 +113,7 @@ impl FontCache {
             ),
         ];
 
-        if let Some(font_ctx) = font_context.inner.try_read() {
+        if let Some(mut font_ctx) = font_context.inner.try_write() {
             for &ch in &common_chars {
                 for &attrs in &common_attrs {
                     let key = (ch, attrs);
--- a/frontends/rioterm/src/renderer/mod.rs
+++ b/frontends/rioterm/src/renderer/mod.rs
@@ -418,7 +418,7 @@ impl Renderer {
 
         // Batch font lookups with a single lock acquisition
         if !font_lookups.is_empty() {
-            let font_ctx = self.font_context.inner.read();
+            let mut font_ctx = self.font_context.inner.write();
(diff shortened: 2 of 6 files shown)


raphamorim:
Hi @itsjunetime thanks for the PR!!! I super appreciate. This I might need to think a bit more, Rio used to store all fonts but some fonts, has data that spikes a lot the memory usage specially if they aren't being used, that being said also write rwlock is costful to the renderer (although in this case we can fix by making a check if the font data exists)

This case also, the write locks can't be removed because we save the data already in the font. So is basically always read, unless fonts are changed.

But yea it does trashing rn. Other idea is to load the font in background/non-blocking, need to think a bit more on this case.

raphamorim:
Also if you have discord feel free to join to discuss or ping me directly as well. 👍 

Might gonna take next days to look how other terminals/editors do as well

raphamorim:
For now I added a more simple fix for this issue a7be7cb4d3a7f1acbdeb37d9e4420600f4b0bf1c added you as co-committer since came from ideas of this PR. Could you check if helps?

itsjunetime:
That commit seems to completely get rid of the performance issues I was seeing, thanks!

PR #1282, opened 2025-08-21
Don't copy data as much; use more SharedData

itsjunetime:
This fixes the issues I described in #1281 - the commit made to fix this helped somewhat, but only avoided the I/O - there was still a lot of memcpy going on. So this just avoids all those memcpys and makes everything a SharedData instead. As far as I can tell, this will only decrease memory usage and improve performance.

### DIFF
--- a/sugarloaf/src/font/mod.rs
+++ b/sugarloaf/src/font/mod.rs
@@ -22,7 +22,7 @@ use crate::font_introspector::text::Script;
 use crate::font_introspector::{CacheKey, FontRef, Synthesis};
 use crate::layout::FragmentStyle;
 use crate::SugarloafErrors;
-use parking_lot::{Mutex, RwLock};
+use parking_lot::RwLock;
 use rustc_hash::FxHashMap;
 use std::ops::Range;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


sugarloaf/src/font/mod.rs
```diff
@@ -682,11 +674,7 @@ impl FontData {
         let stretch = attributes.stretch();
         let synth = attributes.synthesize(attributes);
 
-        let data = if evictable {
-            None
-        } else {
-            Some(SharedData::new(data))
-        };
+        let data = evictable.then_some(data);
```
itsjunetime, line 677: Ah. Oops
```suggestion
        let data = (!evictable).then_some(data);
```

sugarloaf/src/font/mod.rs
```diff
@@ -31,7 +31,7 @@ use std::sync::{Arc, OnceLock};
 pub use crate::font_introspector::{Style, Weight};
 
 // Type alias for the font data cache to improve readability
-type FontDataCache = Arc<Mutex<FxHashMap<PathBuf, Arc<Vec<u8>>>>>;
+type FontDataCache = Arc<RwLock<FxHashMap<PathBuf, SharedData>>>;
```
raphamorim, line 34: You chose RwLock because we write less than read right? asking by curiosity

PR #1283, opened 2025-08-21
Pr 1282

raphamorim:


raphamorim:
@itsjunetime what you think? found this project that has a good implementation of RwLock<HashMap> https://github.com/xacrimon/dashmap

raphamorim:
I didn't know https://github.com/willothy/whirlwind seems pretty cool as well.

sugarloaf/src/font/mod.rs
```diff
@@ -536,14 +536,7 @@ impl SharedData {
...
         Self {
-            inner: Arc::from(data),
-        }
-    }
-
-    /// Creates shared data from an Arc<Vec<u8>> (avoids cloning)
-    pub fn from_arc(data: Arc<Vec<u8>>) -> Self {
-        Self {
-            inner: Arc::from(data.as_ref().clone()),
+            inner: Arc::from(data.into_boxed_slice()),
```
itsjunetime, line 539:
```suggestion
            inner: Arc::from(data),
```

std has [an impl](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-From%3CVec%3CT,+A%3E%3E-for-Arc%3C%5BT%5D,+A%3E) so you don't have to make an intermediate boxed slice (though I doubt this'll actually affect codegen at all, it's just easier to read imo)

sugarloaf/src/font/mod.rs
```diff
@@ -906,25 +895,26 @@ fn find_font_path(
...
-            // Store in cache for future use
-            let font_data_arc = Arc::new(font_data);
-            cache_guard.insert(path.clone(), Arc::clone(&font_data_arc));
-            return Some(font_data_arc);
+            let shared_data = SharedData::new(font_data);
+            // Use entry API to handle concurrent inserts properly
+            let entry = cache
+                .entry(path.clone())
+                .or_insert_with(|| shared_data.clone());
+            return Some(entry.clone());
```
itsjunetime, line 917: This makes sense to me. I was gonna suggest just inserting since that guarantees you're getting the most recent version of the data, but this allows you to potentially avoid a write to the map, which is nice. So I'm chill with this.

itsjunetime (approved):
👍

sugarloaf/src/font/mod.rs
raphamorim: Yea, just trying to avoid write as possible but frankly as we talked idk even if it will be like oh so much is happening 😆

PR #1289, opened 2025-08-28
Avoid hint label collision

edralston:
If a hint pattern results in the number of hint labels exceeding the size of the alphabet, then some of the labels will need to be two characters instead of just one character.  Sometimes this results in a mixture of single-character and two-character hint labels.  However, as currently implemented, there is a chance for some of the two-character hint labels starting with the same character as a single-character label.  For example, the set of hint labels might contain "j" and "ja" and "jb" and "jc" but in this case, it would be impossible to select the object associated with hint label "j" since the hint engine will still be attempting to match against "ja" or "jb" etc.

This commit resolves this issue by removing the shorter label ("j" in the example above) before adding the longer labels ("ja" and "jb" etc.).


raphamorim:
man this was such good pr i apologise, sorry for late reply.

I will review it if is still interest @edralston

PR #1309, opened 2025-09-24
fix kitty keyboard mode stack logic

aymanbagabas:
This limits the keyboard mode stack depth to 8, which should be more than enough for any reasonable use case. It also changes the stack from a Vec to a fixed-size array, which should reduce allocations and improve performance.

Report now correctly reports the current keyboard mode from the stack. Push and pop operations now correctly manage the stack index and wrap around when necessary.

### DIFF
--- a/rio-window/src/platform_impl/macos/app_delegate.rs
+++ b/rio-window/src/platform_impl/macos/app_delegate.rs
@@ -33,6 +33,8 @@ use crate::event::{
 use crate::event_loop::{ActiveEventLoop as RootActiveEventLoop, ControlFlow};
 use crate::window::WindowId as RootWindowId;
 
+pub const UTF8_ENCODING: usize = 4;
+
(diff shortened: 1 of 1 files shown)


rio-backend/src/crosswords/mod.rs
```diff
@@ -1193,13 +1195,9 @@ impl<U: EventListener> Crosswords<U> {
...
             &mut self.inactive_keyboard_mode_stack,
         );
-        self.set_keyboard_mode(
-            self.keyboard_mode_stack
-                .last()
-                .copied()
-                .unwrap_or(KeyboardModes::NO_MODE)
-                .into(),
-            KeyboardModesApplyBehavior::Replace,
+        mem::swap(
```
raphamorim, line 1198: 👌

raphamorim (approved):
👌

PR #1362, opened 2025-11-13
 Implement `--app-id` flag

pinpox:
This adds the `--app-id` flag on Linux, similar to other terminal emulators (e.g. foot), which allows setting the wayland app or X11 WM_CLASS property using `--app-id=mytermimal`.

 It is useful in window managers like sway, that allow to defind custom rules for specific windows, e.g. to implement a "dropdown" terminal.


raphamorim (approved):
thank you!

PR #1368, opened 2025-11-23
Attempted Fix for Input Hanging on Wayland

### LINKED ISSUE #1081: Input freezes / hangs when no input for a few seconds
When I don't input anything into Rio for a few seconds (~10sec), input will freeze / hang.  When I then start typing again, I won't see my typing reflected in the terminal for several seconds (1-3sec).  After a few seconds I can type and work normally again, with my input reflecting immediately, until I stop typing again for a few seconds (usually when I switch windows).

This seems to be input-related as opposed to be output-related: `seq 1 100 | each {date now | print ; sleep 1sec}` (Nushell) will continue printing lines, and I will see my input reflected immediately if I return focus to the Rio window and start typing.  But when I leave the window alone for just 10sec and then return, I will not see what I type for 1-3sec.

This does not only affect the shell prompt, but also occurs when I have programs like Nano or Meli open in Rio: For several seconds I will also not see text I enter when returning to the Rio window.

Other terminal emulators, like Alacritty, appear to be unaffected.

The problem has been ongoing for weeks or months, i.e. it is not a recent regression.

I am using Rio 0.2.12 on NixOS unstable with Mesa 25.0.4.

```
Operating System: NixOS 25.05
KDE Plasma Version: 6.3.4
KDE Frameworks Version: 6.13.0
Qt Version: 6.9.0
Kernel Version: 6.14.3 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 9 6900HS with Radeon Graphics
Memory: 30,6 GiB of RAM
Graphics Processor 1: AMD Radeon 680M
Graphics Processor 2: AMD Radeon RX 6800S
Manufacturer: ASUSTeK COMPUTER INC.
Product Name: ROG Zephyrus G14 GA402RK_GA402RK
System Version: 1.0
```

```
 - system: `"x86_64-linux"`
 - host os: `Linux 6.14.3, NixOS, 25.05 (Warbler), 25.05.20250423.8a2f738`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.28.2`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/store/2ab9ljvsgkmn3zizflwf61207a23q6q1-source`
```

jayhwkns:
Intended to fix #1081 but it doesn't quite work. Removed some dead code and moved Wayland's event dispatcher to an Rc instead of cloning twice like before.

- Added (and commented out) a print statement that helps mitigate the issue #1081 
  - This a a *heisenbug* of sorts. Routinely calling a print statement (in this case in the title update function) mitigates the bug, but only if running from a terminal and printing to stdout (not `/dev/null` or any other file).
- Removed unused function prototype for `user_event`. Unrelated to the issue, but just something that I'd notice.
- Moved some references to the Wayland event dispatcher to an Rc to prevent cloning.
  - This was intended to be a solution, as I was told (by ChatGPT so take with a grain of salt) that cloning the dispatcher did something to the file descriptors that messed up polling. Still, this should save on resources slightly.

I'd be happy if this got merged, but I mostly intend for this to be a starting point for solving #1081 as it's affecting me. Highly curious about what the actual solution is.

### DIFF
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -706,6 +706,7 @@ impl<T: EventListener + Clone + std::marker::Send + 'static> ContextManager<T> {
                     self.titles.set_key_val(i, content, None);
                 }
             }
+            //println!("MITIGATION");
 
             self.titles.set_key(id);
         }
--- a/rio-window/src/application.rs
+++ b/rio-window/src/application.rs
@@ -86,9 +86,7 @@ pub trait ApplicationHandler<T: 'static = ()> {
     /// Emitted when an event is sent from [`EventLoopProxy::send_event`].
     ///
     /// [`EventLoopProxy::send_event`]: crate::event_loop::EventLoopProxy::send_event
-    fn user_event(&mut self, event_loop: &ActiveEventLoop, event: T) {
-        let _ = (event_loop, event);
-    }
+    fn user_event(&mut self, event_loop: &ActiveEventLoop, event: T);
 
     /// Emitted when the OS sends an event to a winit window.
     fn window_event(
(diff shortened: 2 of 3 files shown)


raphamorim:
Sorry took forever but fixed in https://github.com/raphamorim/rio/commit/f7f842d42b6addd603f62360fa88d79ffb5c4255

PR #1402, opened 2025-12-29
Add set-min-size configuration option

TheZoq2:
This is a bit of an odd feature, but I like to have my terminals squished vertically for long running tasks

image

However, `niri` does not squish windows if they set a minimum size which this addresses.

I was considering allowing a setting for the minimum size too, but since there is no canonical way to set `None` in `toml` I figured this is the least controversial way I could come up with

### DIFF
--- a/docs/docs/config.md
+++ b/docs/docs/config.md
@@ -1581,6 +1581,21 @@ Example:
 height = 400
 ```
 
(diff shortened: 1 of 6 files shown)


raphamorim (approved):
looks good, just a small comment to keep pattern with other configs

PR #1420, opened 2026-01-16
[warnings] Disabled warnings

JRBros2346:
Used `Clippy` as baseline

Fixed a bug where a local variable is modified instead of modifying a reference.

Known Issue:
- Only includes Windows, as I am using Windows System


JRBros2346:
Looked into the logs of the all the checks. Looks like there is no warnings in all of the different environments.

brian6932:
Would be nice if this was merged. There's a whole lot of warning text when building now, a lot more than even this PR covers.

PR #1434, opened 2026-02-03
Fix IME preedit overlay and add tests

Tryanks:
before

before

after

after


JRBros2346:
Run `cargo fmt` and push changes. It must make the Checks to pass

u1-liquid:
The current CI failure appears to be caused by an external factor rather than this change.
If this fix is merged, it will significantly improve usability for CJK users, so it would be very helpful if the priority could be raised.

At the moment, the IME pre-edit text is not rendered at all, and users sometimes have to rely on a popup that is completely misaligned to input text. This makes typing quite inconvenient for CJK users.

raphamorim:
Is PR still valid? if so, I would like to help going out

Tryanks:
> Is PR still valid? if so, I would like to help going out

Fixed, but need more testing for CJK, not only chinese

shiena:
I tested this on Windows with Microsoft IME and can confirm that Japanese preedit text (composition string) is now displayed correctly in Rio. This is a great improvement for CJK input users. 🎉

https://github.com/user-attachments/assets/1005f42a-d64f-410d-aec3-d1ff367ecd0b

One thing I noticed: this branch appears to have conflicts with the current `main` branch. Could you rebase onto `main` and resolve the conflicts? That should make it easier to review and merge.

shiena:
Apologies for jumping in from the side, @Tryanks — I needed this fix
locally, so I rebased your branch onto current `main` and ended up
tweaking a few things on top for my own use. Sharing in case any of it
is useful to you or @raphamorim; feel free to cherry-pick, ignore, or
point out anything that doesn't match your intent.

Branch: https://github.com/shiena/rio/tree/improve/ime-preedit-overlay

- [`d579baf0`][d579baf0] — your patch rebased onto current `main`, conflicts resolved. No behavior change.
- [`fa21e7601d`][fa21e76] — the local tweaks described below.

[d579baf0]: https://github.com/shiena/rio/commit/d579baf06dd91faba05ea6482242ca32d23ce52f
[fa21e76]: https://github.com/shiena/rio/commit/fa21e7601ddd5a00630d76c6d7eca744644227e1

### Notes on `fa21e7601d`

While using the rebased branch I ran into a few things that nudged me
into experimenting further. These are just what worked for my setup
(Japanese IME on Windows), not necessarily the right call for the PR:

- **IME cursor tracking inside the preedit.** `PreeditOverlay` picks up
  `Preedit::cursor_byte_offset` and records an `ime_cursor_pos` by
  walking the text in `ch.len_utf8()` steps. This let me render the IME
  caret when arrow-key navigating inside the composition — without it I
  couldn't tell where the IME cursor was.

- **Composing past empty cells.** I was losing preedit characters on
  cells whose underlying `square_content` is `\0` (typing past
  end-of-line). Moving the preedit override *before* the `\0`
  early-continue fixed it. Wide spacer slots (both `Wide::Spacer` and
  the overlay's `PreeditCell::Spacer`) are skipped so wide glyphs don't
  get pushed wider.

- **Reverse-video block rendering.** With my CJK input the
  `DimForeground` + per-cell underline treatment showed gaps between
  characters and only the first cell was inverted, so the composition
  was hard to read at a glance. I tried painting every composition cell
  as a block cursor (wezterm does something similar) and it read much
  better as a single region. I kept a beam on top for the IME caret so
  navigation inside the composition is still visible.

- **Plain font attributes on composition cells.** My prompt has italic
  segments (Starship), and the preedit was inheriting them. Forcing
  `Normal` stretch / weight / style stopped the composing text from
  being italicised.

- **A few more `PreeditOverlay` tests** covering IME cursor at end,
  inside the preedit, `None` offset, and the start-of-preedit case.

The reverse-video change is the most opinionated one and definitely
deviates from the original direction of this PR, so please take it as a
reference rather than a proposal — I'm fine leaving it on my fork if
it doesn't fit.

Tryanks:
> Apologies for jumping in from the side

@shiena Feel free to submit your PR. I’m not sure if I’ll still have time to keep following up with upstream until it gets merged, but I’ll do my best :)

raphamorim:
i can take up the changes and finish in other pr if is ok, btw one question. this never worked well then in alacritty for example right?

shiena:
Hi @Tryanks — while testing this overlay I found a related issue that still reproduces on top of this PR. Flagging it here; happy to open a follow-up PR if useful.

**Problem**

Committing a composition → jumping to the line head with `Ctrl-A` / `Home` → immediately starting the next composition makes the new preedit (and its reverse-video block) briefly render at the **previously-confirmed line end** instead
of the current cursor. Root cause: `terminal.cursor()` is updated asynchronously by the PTY thread, so when the IME preedit event fires before the shell's commit echo / Home response has been parsed, the overlay is placed at a stale
position.

**Reproduction (Microsoft IME on Windows)**

Preconditions:
- Microsoft IME, Hiragana + Romaji input mode
- Debug build of rio, inside a readline shell (WSL / msys bash / zsh). A heavy prompt (`starship`, `oh-my-posh`) or an ssh session makes the race essentially deterministic.
- IME initially off (ASCII mode)

> [!IMPORTANT]
> Timing is the whole point of this bug. The keys in step 2 must be hammered out as a **single rapid burst** with ≤ ~50 ms between each keystroke. If you pause anywhere, the PTY catches up and the race disappears. Longer PTY latency
(WSL / ssh / heavy prompts) widens the window dramatically.

Steps:
1. Press `半角/全角` to enable the IME.
2. **In one uninterrupted burst**, tap in order:
   - `a` `i` `u` — preedit `あいう`
   - `Enter` — commits; the cursor jumps to col 6
   - `Ctrl-A` (or `Home`) — jump to the line head
   - `k` `o` — start the next composition, preedit `こ`

Expected: `こ` appears at col 0.
Actual: `こ` (and the reverse-video block) lands around col 6 — the previous line end — and only snaps back to col 0 a frame or two later when the PTY catches up.

**Fix**

On the first frame of each composition, capture the cursor position the user most recently saw (`renderable_content.cursor.state.pos` before it gets overwritten from the snapshot) and pin the overlay to it until the preedit clears. That
 way the preedit stays where the user intended to type regardless of PTY lag.

Patch on my fork: shiena/rio@d56a5dc0b2f081b08d52efa217abdcaaefc97d4d

PR #1493, opened 2026-04-10
Fix kevent compilation on FreeBSD

rudrabhoj:
Rio doesn't compile on FreeBSD because the `libc::kevent` struct has an `ext: [u64; 4]` field that the `kevent!` macro doesn't initialize.

This adds a FreeBSD-specific variant of the macro that includes `ext: [0; 4]`. The non-FreeBSD variant is unchanged so macOS, DragonFlyBSD, etc. are not affected.

Tested on FreeBSD 15.0-RELEASE with libc 0.2.174.


raphamorim (approved):
Thank you!

PR #1501, opened 2026-04-12
Fix macOS Metal window transparency

### LINKED ISSUE #1500: macOS Metal transparency stays opaque in 0.3.x
## Summary
On macOS, Rios direct Metal backend can render fully opaque even when `window.opacity < 1.0` and blur is enabled.

## Reproduction
Use a minimal config such as:

```toml
[window]
opacity = 0.82
blur = true
decorations = "Transparent"
```

Restart Rio on macOS.

## Expected
The window background should be transparent / blurred according to `window.opacity`.

## Actual
The window stays visually opaque.

## Notes
This appears specific to the direct `Metal` path in 0.3.x rather than config parsing.
The regression is reproducible with the default macOS backend and does not require `WgpuMetal`.

I have a fix ready and will open a PR that preserves alpha in the Metal path and avoids clearing the Metal drawable with a semi-transparent terminal background on transparent macOS windows.

vcali:
Closes #1500.

## Summary
- preserve alpha on the direct macOS Metal path instead of forcing the drawable/window background back to opaque
- let transparent Metal windows use the Cocoa window background tint instead of clearing the Metal drawable with a semi-transparent terminal background
- keep the custom macOS view/layer non-opaque so AppKit/Core Animation can composite the window correctly
- add focused renderer tests for the new transparency decision logic

## Root cause
The 0.3.x macOS Metal path could still end up visually opaque even with `window.opacity < 1.0` because background updates restored alpha to `1.0`, and the renderer was still clearing the Metal drawable directly for transparent windows.

## Testing
- cargo test -p rioterm


raphamorim:
I apologise taking time to reply, will look this soon

vcali:
I'll fix the conflicts later today

corcovado/test/mod.rs
```diff
@@ -12,10 +12,7 @@ extern crate fuchsia_zircon as zircon;
 
 pub use ports::localhost;
 
-mod test_close_on_drop;
```
vcali: @raphamorim looks like the previous PR https://github.com/raphamorim/rio/pull/1535 that cleaned up some tests broke the builds as some of the modules that were deleted were still being required. I'm also fixing it as part of this PR

aekasitt:
Installed via `cargo` as such
```
cargo install --git https://github.com/raphamorim/rio.git --rev refs/pull/1501/head
```
Works !!

---
Workstation details: **Apple M1 Pro (Tahoe 26.3)**

lazo4:
It worked for me a few days ago, but when I try it now, I get: 
Capture d’écran 2026-04-26 à 16 58 11
Notes: I have an M1 MacBook Air, running macos 26.4

vcali:
@lazo4 can you get me your config please?

lazo4:
Sure, I reduced it to the minimal config: 
```
[window]
opacity = 0.5
blur = true
```
All the other options have been left to their defaults

## Important Note
It used to work a few weeks ago when I first tried it, so it’s possible something broke in `main`.
I also have this issue on my 2019 Intel imac, which is running macos 15.7.4.

raphamorim:
Should be fixed in https://github.com/raphamorim/rio/pull/1576 thank you !

PR #1517, opened 2026-04-15
Fix terminal origin jump when switching tabs

### LINKED ISSUE #1513: Strange offset at top left for some tabs
```
OS: Arch Linux x86_64
Kernel: Linux 6.19.11-arch1-1
WM: Sway 1.11 (Wayland)
Shell: bash 5.3.9
Terminal: rio 0.3.10
```

https://github.com/user-attachments/assets/7cbf4947-9866-47a0-8ee0-17c44516fcd1

0x0aa:
## Summary

Fixes #1513 — with multiple tabs open, the terminal content in different tabs could land at slightly different top-left origins, so switching between tabs produced a visible jump.

Reproduced on Arch Linux / Sway (Wayland) at display scale 1.0. The affected code paths in `layout/mod.rs` and `screen/mod.rs` are platform-independent, so the fix should apply equally across platforms.

## The bug

`ContextGrid::apply_taffy_layout` positioned every panel using Taffy's `layout_rect`:

```rust
let [abs_x, abs_y, width, height] = item.layout_rect;
let x = (abs_x + self.scaled_margin.left) / scale;
let y = (abs_y + self.scaled_margin.top) / scale;
```

For a single-panel grid that rect isn't `(0, 0)` — Taffy gives the panel node a non-zero `location` based on the panel's own margin/padding, and the root container's gap. That offset then got added to `scaled_margin` on top of the window margin, so single-panel tabs ended up shifted away from the origin used when the first tab is set up in `Screen::new` (which writes `renderer.margin` directly to `sugarloaf.set_position`). The offset was small and only visible when flipping between tabs.

A second, narrower mismatch existed in `Screen::create_tab`: it fed `current_grid().scaled_margin.left` (physical pixels) into `sugarloaf.set_position`, which expects logical pixels. At scale 1.0 the two happen to agree, but on HiDPI displays the new tab landed at a different x than tab 1.

## The fix

- `layout/mod.rs`: in `apply_taffy_layout`, when there's only one panel, bypass Taffy's `layout_rect` and use `(0, 0)` with the full available area minus the window margin. This is what actually resolves the cross-tab jump. Multi-panel grids still use Taffy's rect — they need it for split positioning and clipping.
- `screen/mod.rs`: `create_tab` now uses `self.renderer.margin.left` (logical margin), matching the position set for the first tab in `Screen::new`. This aligns the initial rich-text position with tab 1 on HiDPI.

## Test plan

- [ ] With multiple tabs open (no splits): switch between them — the top-left of the terminal content stays fixed, no jump across tabs.
- [ ] With non-zero panel `margin`/`padding` in config: same scenario — origin still stays fixed across tabs.
- [ ] With a split pane and multiple tabs: switch between a single-panel tab and a split tab — both share the same top-left origin; split borders and clipping look unchanged.


0x0aa:
Fixed in main

PR #1538, opened 2026-04-19
fix linux x11 1x CPU stuck 100% issue

Buhrietoe:
Found that in 0.3.11 introduced a bug where 1x cpu was at 100% usage at idle with default configuration.

Not familiar with the project but local qwen3-coder-next-80b helped me pin this down and this works for me.


raphamorim:
Thanks for the pr! This should be fixed on #1553 .

PR #1541, opened 2026-04-20
Glyph protocol

raphamorim:
Ref: https://rapha.land/adding-color-glyphs-to-glyph-terminal-protocol/
Ref: https://rapha.land/introducing-glyph-protocol-for-terminals/


raphamorim:
Closing this in favor of #1566

PR #1547, opened 2026-04-21
fix: centre wide-slot glyphs using the shaper advance

### LINKED ISSUE #1546: Wide CJK / emoji glyphs overlap the next cell (regression in main since 0.3.11)
## Summary

Characters with natively-wide glyphs (CJK ideographs, emoji, fullwidth ASCII) visually overlap the next cell when followed by a narrow character. The overlap is not present in `v0.3.11` but reproduces consistently on current `main`.

## Reproduction

```bash
echo "😀A"
```

Works with any emoji (color or monochrome) or CJK codepoint. Variants for environments without emoji
fallback:

```bash
echo "案A"
printf '\xe6\xa1\x88A\n'
```

## Expected vs Actual

- **v0.3.11** (expected behaviour)
    
    Image
- **main / nightly** (`rioterm 0.3.12` at `60185672bf`) — `A` overlaps the right edge of the preceding
glyph
  
    Image

## Environment

- Affected: `rioterm 0.3.12` from `main` at commit `60185672bf`
- Not affected: `rioterm 0.3.11` (release)
- OS: Windows 11
- Font: [UDEV Gothic](https://github.com/yuru7/udev-gothic)

shiena:
fixed #1546

`cell_width * (char_width - 1.0) / 2.0` assumed every glyph's shaper advance was one cell. For natively wide glyphs (CJK, emoji, fullwidth ASCII) the advance is already two cells, so the +0.5 cell shift pushed them into the right neighbour.

Use `(effective_advance - shaper_x_advance) / 2` so narrow-in-wide slots still centre and wide-in-wide slots stay at the left edge.


shiena:
The underlying issue (#1546) no longer reproduces on current `main` ( 105e485c17013b0dac510b00bf4e3a2af42cf6b3 ), likely fixed by the recent font-system rework ( 706a2fdf34 and surrounding commits).

Closing as the standalone fix is no longer needed.

Issue #1528, opened 2026-04-17
Creating a new tab doesn't update the height of the console


NSPC911:
https://github.com/user-attachments/assets/77125d40-cdc1-461a-a1f0-6899c15701cb

also can you bring back bookmarks

raphamorim:
Fixed by #1723 and will be part of the 0.4.12 release. Root cause was the same stale layout size described in #1495: the new tab bar margin never triggered a real layout pass, so the terminal kept its old height until a window resize. Retesting on 0.4.12 would be appreciated.

PR #1552, opened 2026-04-23
fix: bottom row clipped and input frozen when the tab bar appears

### LINKED ISSUE #1495: on opening new tab the bottom of the terminal is hidden until window resize
When opening a new tab the bottom part of the terminal goes out of view. Specifically (at least with my font / font size) the last line is cut in half.
After closing the first tab (so the tab bar disappears) the view shifts upwards so the last line is now too far from the edge of the window. 
Resizing the window at any point fixes the problem

After new tab:
Image

After closing initial tab:
Image

On linux (EndeavourOS) - gnome - rioterm 0.3.4

shiena:
Fixes #1495, #1528.

When `navigation.hide_if_single = true` (the default) and a second tab is opened, the tab bar appears and shifts the render origin down by the island height (~34 px).
Before this change, the terminal panels were still sized against the full window height, so the last row ended up drawn past the bottom of the visible window area.
With a horizontal split, the bottom pane's last row disappeared entirely.

A follow-up symptom showed up once the layout was corrected: after closing back to a single tab (via shell exit), keystrokes no longer produced any visible output until the window was clicked or a tab-switch shortcut forced a render. This turned out to be a pre-existing damage-routing bug that the layout fix surfaced.

Combined with #1518 this fully resolves #1495: this PR handles the "new tab → bottom clipped" direction and the stuck-input regression, while #1518 covers the "shell exit → top strip retained" direction.


shiena:
Rebased only the minimum set of commits required to apply this PR onto the current main (97340fe5410cf8324feb27ce81c4066e4fd09750).

raphamorim:
The margin/height propagation this fixes was rebuilt in #1723 and both #1495 and #1528 are closed, so closing as superseded. The repaint-on-shell-exit half is being handled in #1688. Thanks!

PR #1557, opened 2026-04-25
fix(ime): render preedit inline and stop Windows key-repeat from stalling redraws

### LINKED ISSUE #1252: macOS: Cannot type in Japanese
Running on macOS Sequoia. With the Japanese Roomaji input method enabled, text does not appear until the return key is pressed. It should show preview text and allow the user to select the correct word/spelling.

shiena:
- **Follow-up to #1434**: the v4 grid-renderer refactor moved emission into `grid_emit.rs`, so #1434's per-cell
`create_style` patch no longer applies. Re-implemented against the new path: `PreeditOverlay` is threaded through
`build_row_bg` / `build_row_fg`, composition cells render as a wezterm-style block (cursor color bg, glyph inverted),
Spacers skip emission so wide CJK chars stay contiguous, and the IME caret is a new `DecorationStyle::ImeCaret` sprite.
Co-Authored-By Tryanks.
- **Render during IME key-repeat (Windows)**: sustained IME repeat floods the message queue with `WM_KEYDOWN` + `WM_IME_*`
 + `TerminalDamaged`, starving `about_to_wait` so the scheduler never fires a paint. Added a custom `REDRAW_REQUESTED_MSG`
 posted by `Window::request_redraw` — posted messages return from `PeekMessageW` ahead of `WM_PAINT`, so paints interleave
 with the flood. The handler clears the DwmFlush worker's dirty flag to avoid double-painting.
- **Speed up IME key-repeat (Windows)**: `WM_IME_STARTCOMPOSITION` / `ENDCOMPOSITION` were forwarded to `DefWindowProc`,
which runs a ~100 ms synchronous TIP handshake — capping corvus-skk direct-input at ~10 cps. Returning 0 (wezterm already
does this for END) drops the per-keystroke cycle to ~31 ms, matching the OS repeat rate.


shiena:
Rebased onto #1849. Two commits are gone as a result:

- `Render IME preedit as wezterm-style block in the grid renderer` — superseded
  by `PreeditLine`. Every conflict hunk was a case of #1849 carrying the later
  design, so there was nothing left to port.
- `[ime] fix: follow the live cursor for IME preedit instead of pinning` — this
  dropped anchoring machinery that #1849 never introduced; it already rebuilds
  the line from the snapshot cursor each frame.

PR #1558, opened 2026-04-25
fix: respect hide_if_single across layout and input handlers

shiena:
Follow-up to #1518 (extending #1516), rebased onto `main` to resolve conflicts. Adds further fixes uncovered while testing the original
change:

- **Keep tab-bar margin transitions in sync with `hide_if_single`** (110b1c356a) — `resize_top_or_bottom_line` compared
the new padding against `dimension.margin`, which `apply_taffy_layout` zeroes on every run, so the `hide_if_single`
transition was missed and `update_scaled_margin` never fired. Now compares against the grid's authoritative
`scaled_margin`, and `update_scaled_margin` also resyncs Taffy's root size so panels reflow into the freed area. Without
this, closing the second tab left the tab-bar strip reserved at the top; spawning a new tab pushed the bottom row past the
 window.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1362,13 +1362,16 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     }
                 }
 
-                // Check if mouse is over island and set cursor to default
+                // Check if mouse is over island and set cursor to default.
+                // Skip guard when island is hidden (hide_if_single + single tab).
                 use crate::renderer::island::ISLAND_HEIGHT;
                 let scale_factor = route.window.screen.sugarloaf.scale_factor();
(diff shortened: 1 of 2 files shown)


shiena:
Merging only one of them leaves a different hide-if-single symptom in
place:

- With only #1552: closing the second tab with hide-if-single enabled
  leaves the tab-bar strip reserved at the top of the window, so
  terminal content draws below it instead of filling the freed area.
- With only #1558: toggling the tab bar (via tab open/close under
  hide-if-single) leaves split panels sized for the previous margin,
  and the bottom row of the terminal lands outside the visible window.

raphamorim:
Superseded by #1723: resize_top_or_bottom_line now compares against the grid's scaled_margin and re-lays-out every grid, the approach this PR proposed. Thanks!

PR #1581, opened 2026-05-03
ci: drop --release from cargo test to avoid OOM on ubuntu runner

shiena:
The `test-native (ubuntu-latest)` job has been failing with `exit code 143` (SIGTERM) on every PR. The runner is killed during the link step of `rioterm` due to memory exhaustion, not a CI timeout.

This drops `--release` from the test command so `cargo test --features wgpu` runs in the dev profile, which compiles with parallel codegen and no LTO.


raphamorim:
It was debug mode before, I don't remember why I went to --release but I think is fair to go back. Thank you!

PR #1583, opened 2026-05-03
fix: window.opacity transparency on Windows (wgpu + cpu backends)

shiena:
## Problem

On Windows, `[window] opacity < 1` rendered the window almost entirely see-through instead of partially translucent.
Reproducible with the default `cargo build` (wgpu backend) and with `[renderer] use-cpu = true` (cpu backend).

## Root cause

`with_transparent(opacity < 1.)` triggers `DwmEnableBlurBehindWindow`, after which DWM uses the framebuffer alpha for per-pixel compositing.
Both backends were producing the wrong alpha:

**wgpu backend** had two independent issues:

1. `frontends/rioterm/src/screen/mod.rs` gates the wgpu code path on rioterm's own `wgpu` feature, which is **off** in default `cargo build`.
Default Windows builds therefore fell into the CPU rasterizer, even though `sugarloaf` and `rio-backend` already get `features = ["wgpu"]` via the `[target.'cfg(windows)']` dep override, so the wgpu code is in the binary.
2. `Sugarloaf::render_wgpu` wrote `LoadOp::Clear((bg_r, bg_g, bg_b, opacity))` raw.
wgpu picks `PreMultiplied` alpha mode on Windows (the only non-Opaque option DX12 / Vulkan WSI offer), so DWM read the dark bg as if it were already-multiplied — the implied "straight" RGB came out near-fully transparent.

**cpu backend** wrote `0x00RRGGBB` pixels (`α = 0`) and every blend function masked the alpha byte out (`pack_opaque` omits it, SIMD results post-masked with `& 0x00ff_ffff`).
Harmless on platforms that ignore swap-chain alpha; with DWM reading the alpha byte, every pixel came out fully transparent.

## Fix

**wgpu**:

- Allow the wgpu dispatch on Windows even without rioterm's own feature flag: `cfg(any(feature = "wgpu", target_os = "windows"))`.
- Premultiply the `LoadOp::Clear` color in `render_wgpu` for any alpha mode that isn't `PostMultiplied`.
Promote `WgpuContext::alpha_mode` to `pub` so the renderer can branch on it.

**cpu** — switch the softbuffer framebuffer to **premultiplied RGBA** end-to-end:

- `pack_opaque` writes `α = 0xff` (no-op on OSes that ignore the byte).
- bg fill in `render_cpu` uses `pack_premul((R,G,B)·a, a)`.
- All four blend variants in `renderer/cpu.rs` (scalar SWAR + x4/x8 SIMD const-src + var-src) gain an A-channel computation via the same `(x · inv) / 255` SWAR trick as G, and stop stripping alpha from the source / result.
- The two scalar blend paths in `grid/cpu.rs` and `text.rs` (grid cells and UI text overlays) get the same A-channel fix, with a new `pack_premul` helper next to each `pack_opaque`.
- Caller sites that splatted `(src_premul & 0x00ff_ffff)` and post-masked SIMD results are updated to splat the full premultiplied src and drop the post-mask.

The premultiplied invariant (RGB ≤ A) keeps byte-wise channel adds safe from inter-channel carry.
Default-bg cells continue to emit `(0, 0, 0, 0)`; the `sa == 0` early-out returns `dst` unchanged, so the cleared bg's alpha bleeds through.

## Tested

Windows 11 / NVIDIA RTX 5070, `opacity = 0.6`:

- wgpu (Vulkan, `Bgra8Unorm`, `PreMultiplied` surface) — renders as ~60% bg + ~40% desktop.
- cpu (`use-cpu = true`) — renders translucent; tabs, prompt text, glyph AA edges, command palette overlay all composite correctly.

`opacity = 1.0` (default) unaffected on both paths.

### DIFF
--- a/frontends/rioterm/src/screen/mod.rs
+++ b/frontends/rioterm/src/screen/mod.rs
@@ -152,16 +152,36 @@ impl Screen<'_> {
         let backend = if config.renderer.use_cpu {
             SugarloafBackend::Cpu
         } else {
+            // `feature = "wgpu"` here gates rioterm's own dep feature,
+            // which is OFF in the default `cargo build` profile. On
+            // Windows the wgpu code is still compiled into sugarloaf +
+            // rio-backend via the target-specific dependency override
+            // (`[target.'cfg(windows)'.dependencies.sugarloaf] features
+            // = ["wgpu"]`), so we can — and must — take the wgpu code
+            // path on Windows even when rioterm's own `wgpu` feature
+            // isn't on. Without this OS override, default Windows
+            // builds silently fall back to the CPU rasterizer, which
+            // doesn't write per-pixel alpha and therefore renders the
+            // entire window fully transparent under the
+            // `with_transparent(true)` + `DwmEnableBlurBehindWindow`
+            // path that `window.opacity < 1` engages.
             match config.renderer.backend {
                 // `Backend::Vulkan` from the user config means the
                 // native ash backend on Linux. Other OSes fall through
-                // to the wgpu Vulkan path when the `wgpu` feature is
-                // on; otherwise we degrade to CPU rasterizer.
+                // to the wgpu Vulkan path when wgpu is available;
+                // otherwise we degrade to CPU rasterizer.
                 #[cfg(target_os = "linux")]
                 Backend::Vulkan => SugarloafBackend::Vulkan,
-                #[cfg(all(not(target_os = "linux"), feature = "wgpu"))]
+                #[cfg(all(
+                    not(target_os = "linux"),
+                    any(feature = "wgpu", target_os = "windows")
+                ))]
                 Backend::Vulkan => SugarloafBackend::Wgpu(wgpu::Backends::VULKAN),
-                #[cfg(all(not(target_os = "linux"), not(feature = "wgpu")))]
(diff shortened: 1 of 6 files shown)


cpea2506:
@shiena Could you please update the branch? and @raphamorim could you please take a look?
 
I'm using custom build from this branch recently and it works very well on Windows 11.

```toml
[window]
opacity = 0.85

[renderer]
use-cpu = true
```

image

shiena:
@cpea2506 rebased

raphamorim:
Sorry will look this week

raphamorim:
Thank you, the diagnosis and the premultiplied conversion here are exactly right, and they match what alacritty, kitty, and ghostty all do (premultiplied alpha at the surface, which is DWM's convention). Superseded by #1905, which keeps your commits unchanged and adds the review follow-ups on top: an alpha-capable adapter preference so transparency also works when wgpu would otherwise pick DX12 (whose HWND swapchains only expose Opaque, the same gap as wezterm's WebGPU frontend), an A+G lane pairing that makes the CPU blends' fourth channel free (back to two multiplies per pixel), unit tests locking the alpha invariants, and a wgpu_backend cfg alias that also fixed one missed reload-path gate. Closing in favor of that one.

PR #1585, opened 2026-05-04
fix: mark tab dirty after shell exit

ddidderr:
Closing a tab by exiting its shell removes the context in the CloseTerminal handler, but that path left the newly active tab with no pending update. The PTY performer already sends a follow-up RioEvent::Render, so a redraw is queued. However, Renderer::run still gates each panel on is_dirty or force_full_damage. When neither flag is set, the frame is discarded and the closed tab remains visible until another action dirties the screen.

Mark the new active screen dirty after resizing the tab line. This keeps the shell-exit path aligned with other UI mutation paths without adding another redraw request; the existing Render event supplies that.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -497,8 +497,10 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                             event_loop.exit();
                         }
                     } else {
-                        let size = route.window.screen.context_manager.len();
-                        route.window.screen.resize_top_or_bottom_line(size);
+                        let screen = &mut route.window.screen;
+                        let size = screen.context_manager.len();
+                        screen.resize_top_or_bottom_line(size);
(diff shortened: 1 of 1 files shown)


ddidderr:
In simple words: You can now exit your shell with `exit` or `Ctrl+d` and the tab will immediately close instead of being still shown until you actively click another tab.

ddidderr:
I added a `route.request_redraw();` because with `use-fork = false` my initial fix didn't work. 

If you have a better idea, feel free to tell me / improve the PR yourself.

ddidderr:
> I added a `route.request_redraw();` because with `use-fork = false` my initial fix didn't work.
> 
> If you have a better idea, feel free to tell me / improve the PR yourself.

Sorry, my bad. I was testing something else and forgot to include the fix from this PR and got overexcited and added the redraw which is - after testing again - not necessary.

PR #1588, opened 2026-05-06
fix: fixes screen resolution stale bug

tauil:
This fixes an issue that has been annoying me a bit for quite some time.

When working with external monitors, I often connect and disconnect them depending on the situation. This causes the resolution of the screen to change and, therefore, shows weird font sizing in Rio.

The screenshot below shows 2 versions of Rio. The one with the problem (tag `v0.4.2`) (tiny font-size) and the one with this code fix (regular font size). You can notice as well that the close. minimize and maximize buttons were misaligned in the buggy verion.

Screenshot 2026-05-06 at 11 42 53


raphamorim (approved):
Thank you!

PR #1596, opened 2026-05-12
fix: panic on window resize from cols mismatch

### LINKED ISSUE #1593: When maximizing the windows Rio craches.
I get this:
thread 'main' (25107) panicked at frontends/rioterm/src/grid_emit.rs:965:25:
index out of bounds: the len is 88 but the index is 88
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fish: Job 1, 'rio' terminated by signal SIGABRT (Abandon)

kronberger-droid:
Fixes #1593.

Rio panics on window resize:

```
thread 'main' panicked at frontends/rioterm/src/grid_emit.rs:965:25:
index out of bounds: the len is 88 but the index is 88
```

`PanelFrame` read `cols` / `rows` from the live `ContextDimension`, while
`visible_rows` and `cell_styles` were sized from the snapshot taken under
the terminal lock. A resize landing between the snapshot and the paint
pass left the row loop walking past the row's storage.

Source the dims from `renderable_content.columns` / `screen_lines` instead, which is captured under the same lock as the row data, so the loop bound and
the row Vec agree.

Tested in niri (built using flake): resize cycles no longer panic, no visible rendering issues, might not tested all the edge cases tho.

### DIFF
--- a/frontends/rioterm/src/screen/mod.rs
+++ b/frontends/rioterm/src/screen/mod.rs
@@ -3702,11 +3702,15 @@ impl Screen<'_> {
                 let cursor_color = term_colors
                     [rio_backend::config::colors::NamedColor::Cursor as usize]
                     .unwrap_or(self.renderer.named_colors.cursor);
+                // `dim` can race ahead of the terminal during resize;
+                // the snapshot was captured under the same lock as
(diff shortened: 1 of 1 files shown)


raphamorim:
Thanks!

PR #1599, opened 2026-05-15
feat(scroll): add opt-in smooth pixel scrolling

nmb4:
## Summary
- Adds an opt-in `scroll.smooth` setting for pixel-level trackpad scrolling while keeping the existing line-based behavior as the default.
- Carries fractional scroll offset through the terminal renderer and grid shaders so scroll movement can render between row boundaries.
- Fetches an extra visible row during smooth scrolling to avoid gaps, resets smooth offset on jump-style scroll commands, and clamps offset at scrollback boundaries.

## Validation
- `cargo fmt --check`

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1564,6 +1564,14 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                                 // Reset offset to zero.
                                 route.window.screen.mouse.accumulated_scroll =
                                     Default::default();
+                                if route.window.screen.renderer.smooth_scroll_enabled {
+                                    route
+                                        .window
+                                        .screen
+                                        .renderer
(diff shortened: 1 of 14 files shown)


nmb4:
Disclaimer: fully AI (GLM 5.1 via claude code, for transparency).
like i didn't touch one line of code for this, i was surprised myself when it just worked. it's not perfect, theres a bug that causes split panes of the same tab to all pixel-scroll at the same time, but i'll have it fixed asap. it's disabled by default, set mouse.smooth = true in the config to enable it. but man it feels _good_. especially on macos using the trackpad, mabye i can make mouse scrolling feel better as well next.

nmb4:
https://github.com/user-attachments/assets/1f608d16-46ae-4a7e-a708-6a1f2c616d95

PR #1607, opened 2026-05-25
fix(windows): ignore in-band KKP mode requests to avoid ConPTY ESC loss

ci4ic4:
## Problem

When a child process (e.g. zellij, neovim) sends `\x1b[=1u` to enable
Kitty Keyboard Protocol mode, Rio honours the request and starts
encoding keypresses as KKP sequences (`\x1b[codepoint;modifiersu`).

On Windows, these sequences are written to ConPTY's input pipe.
ConPTY processes the `\x1b[` prefix as the start of a CSI sequence
and **consumes the ESC byte** before the bytes reach the child process.
The child receives mangled output like `[103;5u` instead of
`\x1b[103;5u`, breaking keyboard input entirely.

Example: pressing Ctrl+G inside zellij (which requests KKP on startup)
produces `[103;5u[103;5u[103;5u` in the terminal instead of the
expected action.

## Root cause

`set_keyboard_mode`, `push_keyboard_mode`, and `pop_keyboard_modes` in
`Crosswords` have no platform guard. On Windows, ConPTY is the only PTY
mechanism, and it cannot pass KKP input sequences intact.

## Fix

Make `set_keyboard_mode`, `push_keyboard_mode`, and `pop_keyboard_modes`
no-ops on Windows using `#[cfg(not(target_os = "windows"))]`, so that
in-band KKP mode requests from child processes are silently ignored.

`report_keyboard_mode` is **intentionally left unchanged** — it correctly
reports mode 0 on Windows. Well-behaved applications (neovim, etc.) that
query the keyboard mode before activating KKP will receive the correct
response and fall back to traditional key encoding automatically.

## Testing

Verified on Windows 11 Pro: zellij running inside Rio now receives
correct key sequences (Ctrl+G triggers the expected action rather than
producing literal `[103;5u` output).

## Notes

This is a compile-time guard rather than a runtime config option.
ConPTY's inability to pass KKP input sequences is a fundamental
architectural constraint, not a version-dependent behaviour. If a future
Windows/ConPTY version correctly passes KKP sequences through, this
guard can be revisited.

---
*Branch rebased onto current `main` and verified (build, clippy, `cargo test -p rio-backend`: 545 passed) with assistance from Claude Code. The fix itself predates that and was verified manually against zellij/neovim on Windows 11.*

### DIFF
--- a/rio-backend/src/crosswords/mod.rs
+++ b/rio-backend/src/crosswords/mod.rs
@@ -5710,6 +5710,7 @@ mod tests {
     /// - test_keyboard_mode_reset: Terminal reset behavior on keyboard stack
     /// - test_keyboard_mode_stack_underflow_protection: Stack underflow protection
 
+    #[cfg(not(target_os = "windows"))]
     #[test]
     fn test_keyboard_mode_push_pop() {
         let size = CrosswordsSize::new(10, 10);
@@ -5757,6 +5758,7 @@ mod tests {
         assert_eq!(term.keyboard_mode_stack[1], KeyboardModes::NO_MODE.bits()); // Should be cleared
     }
 
+    #[cfg(not(target_os = "windows"))]
     #[test]
     fn test_keyboard_mode_stack_wraparound() {
         let size = CrosswordsSize::new(10, 10);
@@ -5816,6 +5818,7 @@ mod tests {
         }
     }
 
+    #[cfg(not(target_os = "windows"))]
     #[test]
     fn test_keyboard_mode_set_replace() {
         let size = CrosswordsSize::new(10, 10);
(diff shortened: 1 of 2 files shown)


ci4ic4:
Rebased onto current `main` (conflict-free) and re-verified on Windows 11: clean `cargo fmt`/`clippy` on the changed crate and `cargo test -p rio-backend` green (545 passed). zellij/neovim keyboard input confirmed working.

One thing I'd value your steer on: I know KKP is now the always-on default (#`d37bb0c08`), and this PR deliberately *doesn't* change that on Linux/macOS — it only no-ops the in-band mode setters on Windows, where ConPTY consumes the leading `ESC` so the sequences can't reach the child intact.

If you'd rather keep KKP active everywhere and instead fix it at the ConPTY **write path** (e.g. so the `ESC` survives), I'm happy to rework it in that direction — just let me know which approach you'd prefer to carry.

PR #1609, opened 2026-05-25
Help Vulkan glyph atlas texture grow

### LINKED ISSUE #1608: Characters go missing with too many characters (Vulkan)
First, configure a font that is able to show many CJK and Emoji, I select Unifont, as described in https://github.com/raphamorim/rio/issues/1587#issuecomment-4535934147

```
[fonts]
family = "Unifont"
size = 14.0
symbol-map = [
  { start = "10000", end = "20000", font-family = "Unifont Upper" },
]
```

Then, run [ucs-detect](https://github.com/jquast/ucs-detect/),

```
uvx ucs-detect

# or, if you prefer,
pip install ucs-detect && ucs-detect
```

The final output is corrupted:

Image

This table appears strange, some numbers, and '%' is missing. Almost all capital letters are missing, too.

In the bash prompt that follows, I have typed the full ASCII alphabet, and symbol table, but only letters 'A', 'CDEFG', 'I', [..] are visible, the rest appear as spaces. Within the first few pages of output, codepoints begin to go missing. I suspect that the full Chinese language may not even fit on current Vulkan GPU restraints, most certainly not with all of the possible colors and styles.

I believe this will impact mostly only CJK users, especially as custom font large enough to exceed vulkan atlas texture is required to use CJK, and especially when combined with copious amounts of output in varying colors and styles with Vulkan GPU, only.

jquast:
Closes #1608 

Problem
-------

A very strange problem, running this command:

    uvx ucs-detect

    # or,
    pip install ucs-detect && ucs-detect

See before and after video,

https://github.com/user-attachments/assets/5c319b20-7b3f-46b4-8ba5-6972a9b5706f

Vulkan glyph atlas was a fixed size, when it filled up, new glyphs rendered as blanks. In this example, if I type, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' many letters are missing, '%' is missing and others, because by the time they are first used in this process, there is no atlas space remaining for ASCII letters of this color.

rio-vulkan-gpu-doesnt-grow

Solution
--------

Added `grow()` to `VulkanGlyphAtlas`, matching the Metal and CPU backends. The atlas now doubles to 4096, then 8192, copying old content into the new image via `vkCmdCopyImage2`. Descriptor sets are updated after growth so the shader sees the new image view.

Also added grow-retry to the UI text overlay path (`text.rs`), which had the same gap.

Note
----

An LLM (deepseek) was used to discover the problem and solution. I am not an experienced Rust developer and I have never written any Vulkan code of any kind in any language. Please review these changes very carefully. All of the above words are my own.

### DIFF
--- a/sugarloaf/src/context/vulkan.rs
+++ b/sugarloaf/src/context/vulkan.rs
@@ -136,11 +136,10 @@ pub struct VulkanContext {
     swapchain_loader: khr::swapchain::Device,
 
     // Core device.
-    queue: vk::Queue,
+    pub(crate) queue: vk::Queue,
     // Kept around so future phases (atlas uploads, a dedicated transfer
     // pool, pipeline creation) don't have to re-probe the family.
-    #[allow(dead_code)]
-    queue_family_index: u32,
+    pub(crate) queue_family_index: u32,
 
     /// Reference-counted owner of `device`, `instance`, `physical_device`,
     /// and the loader (`Entry`). Cloned into every per-resource struct
@@ -803,7 +802,7 @@ pub fn allocate_host_visible_buffer_raw(
 /// produce for the standard `HOST_VISIBLE | HOST_COHERENT` and
 /// `DEVICE_LOCAL` combinations, but callers should still treat it as
 /// fatal rather than ignore it.
-fn find_memory_type(
+pub(crate) fn find_memory_type(
     instance: &Instance,
     physical_device: vk::PhysicalDevice,
     type_filter: u32,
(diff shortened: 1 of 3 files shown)


jquast:
spirv-cross-sys 0.7.2 fails to compile on MSYS2 Clang after its upgrde from 0.4.3 as a dependency of librashader 0.11, upgraded by a7766425b007d679c2ae4edb512d6420f286d051, it's really outside of my wheelhouse sorry I couldn't help more.

raphamorim:
hey @jquast thanks for the PR!! I will take a look. 🙏 (this week has been quite rushed for me, sorry about delay)

Will run it locally on my linux first, before address changes.

> spirv-cross-sys 0.7.2 fails to compile on MSYS2 Clang after its upgrde from 0.4.3 as a dependency of librashader 0.11, upgraded by https://github.com/raphamorim/rio/commit/a7766425b007d679c2ae4edb512d6420f286d051, it's really outside of my wheelhouse sorry I couldn't help more.

thanks for the heads up

raphamorim:
@jquast could u test later this branch https://github.com/raphamorim/rio/pull/1611 ?

jquast:
@raphamorim yes I can confirm the branch in #1611 this solves the problem described, thanks for your attention.

image

raphamorim:
Our branch 👌 🙏

PR #1613, opened 2026-05-29
feat: ctrl+wheel font-size zoom

### LINKED ISSUE #792: Doesn't zoom in/out on Ctrl + mouse wheel
This is a common feature in terminals, but it doesn't work in Rio.

atreidesend:
Closes #792.

Ctrl+mouse-wheel zooms the font size, matching WezTerm, Windows Terminal, iTerm, and similar terminals. The feature has been requested since November 2024.

Hooks into the existing `MouseWheel::LineDelta` arm: when Ctrl is held (and Shift is not, and the focused app is not in mouse-reporting mode), wheel-up calls `change_font_size(Increase)` and wheel-down calls `Decrease`, reusing the handler the keyboard shortcuts already use. Mouse-reporting mode (`tmux mouse on`, vim `:set mouse=a`) is respected so wheel events still pass through unchanged when the inner app wants them.

Touchpad (`PixelDelta`) input intentionally falls through to the existing scroll path. Touchpad gestures fire many small deltas per swipe, and wiring zoom into them naively gives ~12 zoom steps per swipe; a debounced/threshold-based variant is worth a follow-up once the simpler wheel case is in.

Opt out via:

```toml
[mouse]
wheel-zoom = false
```

Defaults to `true` so the feature is discoverable.

Verified locally on Windows 11.

> Aside: `main` (at `a7766425b0`, "update to wgpu 29.0.3") currently fails to build on Windows MSVC because of a `windows` crate version conflict between `wgpu-hal` 29.0.3 (uses `windows` 0.62) and `gpu_allocator` (still on 0.54) in the dx12 backend. This change was developed and tested against `v0.4.5` and applies cleanly. Happy to file a separate issue for the build break if it'd help.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1,3 +1,4 @@
+use crate::bindings::FontSizeAction;
 use crate::event::{ClickState, EventPayload, EventProxy, RioEvent, RioEventType};
 use crate::ime::Preedit;
 use crate::renderer::utils::update_colors_based_on_theme;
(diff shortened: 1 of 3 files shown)


atreidesend:
Quick note on the red checks here — I don't think they're related to this change. The `msys2-build-test` and `Nix Build` jobs are failing on an MSRV mismatch:

```
error: rustc 1.95.0 is not supported by the following packages:
  rio-notifier@0.4.6 requires rustc 1.96.0
  rioterm@0.4.6 requires rustc 1.96.0
```

It looks like this traces back to `6a11aa33c7` ("update to rust 1.96"), which bumped the workspace `rust-version` to `1.96.0`. The msys2 toolchain currently ships rustc 1.95.0, and the Nix build reports `No stable 1.96.* is available`, so both fail before compiling anything. The same jobs are red on `main` independently of this PR.

The three `test-native` jobs (Windows/macOS/Linux) — which actually build and test the change — are passing. This diff doesn't touch any `Cargo.toml`, dependencies, or the MSRV, so I believe it's unaffected. Happy to rebase once the toolchain situation settles, and of course glad to adjust anything in the change itself.

PR #1614, opened 2026-05-29
Fix Quit action being silently dropped (#892)

### LINKED ISSUE #892: Cannot quit via key binds
On the docs, it says:
> Close tab or quit	Control + Shift + W

Close tab works, except quit doesn't.

Trying the usual Ctrl Q doesn't quit either.

In addition, setting a keybinding to quit as outlined [here](https://raphamorim.io/rio/pt-br/blog/2023/08/02/release-0.0.15#custom-key-bindings) is fruitless, typing v into the terminal

nikicat:
## Summary

The `Quit` action does nothing on Linux/Windows: keyboard binds and the command palette both call `ContextManager::quit()`, which emits `RioEvent::Quit` — but `application.rs` only has a handler for `RioEvent::Exit`, so the event falls through and is silently dropped. (macOS is unaffected because it quits via the `applicationShouldTerminate` path.)

The two events are redundant halves of the same intent:
- `RioEvent::Quit` — sent (`context/mod.rs:673`) but never handled
- `RioEvent::Exit` — handled (`application.rs:422`, with the `confirm_before_quit` logic) but never sent

This consolidates them: the handler now matches `RioEvent::Quit` (consistent with `Action::Quit`, `PaletteAction::Quit`, `ContextManager::quit`, `Route::quit`), and the dead `RioEvent::Exit` variant is removed. The existing handler already respects `confirm_before_quit` and otherwise exits, so no behavior is lost.

Fixes #892.

## Test plan

- [x] `cargo check -p rioterm -p rio-backend` passes with no new warnings
- [x] Confirmed on Linux (Arch/Wayland/GNOME): a custom `[bindings]` Quit entry now quits Rio
- [ ] macOS Cmd+Q still quits (unchanged path)
- [ ] Windows quit binding works

Note: `Action::Quit` only ships in the macOS default bindings, so Linux/Windows users still need a custom bind or the command palette to trigger it — but those paths now work. Adding a cross-platform default keybind is left as a separate decision.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -419,7 +419,7 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     }
                 }
             }
-            RioEventType::Rio(RioEvent::Exit) => {
+            RioEventType::Rio(RioEvent::Quit) => {
                 if let Some(route) = self.router.routes.get_mut(&window_id) {
                     if self.config.confirm_before_quit {
                         route.confirm_quit();
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1774,6 +1774,12 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                                 "yes (y)",
                                 "no (n)",
                             );
+                            // The dialog is an overlay drawn outside the
+                            // terminal's damage tracking. `render()` discards
+                            // the frame when no panel is dirty, which would
+                            // drop the overlay on idle frames — force a
+                            // present so the dialog reliably shows.
+                            route.window.screen.mark_dirty();
                         }
 
                         if let Some(window_update) = route.window.screen.render() {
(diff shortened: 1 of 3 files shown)


nikicat:
Pushed two follow-up commits that fix the quit-**confirmation** dialog (the overlay shown when `confirm_before_quit` is on), which surface now that the Quit action actually fires:

- **fix: always present the quit-confirmation dialog** — the confirm dialog is an overlay drawn outside the terminal's damage tracking. While in `ConfirmQuit` the app runs a continuous redraw loop, but `screen.render()` discards any frame where no panel is dirty, so the overlay only appeared when the terminal happened to be dirty (recent output, cursor blink) and otherwise flickered/was dropped on idle frames. Mark the panel dirty after drawing the dialog so the frame is always presented.
- **fix: repaint when dismissing the quit-confirmation dialog** — pressing `N`/`Esc` set the route back to `Terminal` but never requested a redraw, so the stale dialog overlay stayed on screen until the next event. Request an overlay redraw to repaint immediately, matching the other dismiss paths.

Happy to split these into a separate PR if you'd prefer to keep this one to just the event-dispatch fix.

raphamorim:
The dropped-Quit bug was fixed in 89c5744eff (application.rs now handles RioEvent::Exit | Quit) and #892 is closed. The confirm-dialog follow-ups conflict with the rewritten confirm_quit overlay, so closing this; the dead Exit variant cleanup would be welcome as a tiny separate PR if you want to send it. Thanks!

nikicat:
Opened #1769 with the dead `Exit` variant cleanup you suggested — +1/-5. Thanks for the pointer.

PR #1627, opened 2026-05-30
fix: align URL hint hit-box with wide/multibyte glyphs

### LINKED ISSUE #1619: [Bug] Clickable URL underline/hit-region is horizontally misaligned when a wide (emoji/CJK) glyph precedes the URL on the same line
### Summary

When a plain-text URL is auto-detected, the modifier-hover underline **and the clickable hit-region** are shifted horizontally if any wide (display-width-2) character — emoji, CJK, etc. — appears earlier on the same line. The shift equals the display width Rio under-counts, so the clickable box no longer sits under the visible URL: clicking the URL does nothing, which looks like "the URL isn't clickable."

### Steps to reproduce

Print these four lines, then hold the open-URL modifier (Alt, by default) and hover each:

```
https://example.com/plain      # OK  — underline aligned, clickable
ab https://example.com/ascii   # OK  — ASCII prefix, still aligned
😀 https://example.com/emoji   # BUG — underline shifted right, a click on the URL misses
世界 https://example.com/cjk    # BUG — shifted further (two 2-cell glyphs)
```

### Actual behavior

On the lines preceded by a wide glyph, the underline starts *inside* the URL (e.g. at the 3rd character) and ends the same number of cells *past* the end. Because the underline is the actual click target, clicking on the visible URL text lands in the gap and does nothing. The offset grows with the number/width of preceding wide glyphs (one emoji ≈ +1 cell of error per under-counted glyph; the two CJK chars shift it further).

### Expected behavior

The underline and click-region should align exactly with the URL regardless of any preceding wide glyphs.

### Real-world trigger

Any program that prints a glyph before a URL hits this. For example, Claude Code prefixes a one-line reply with a `⏺ ` bullet; when the reply is just a URL (`⏺ https://linear.app/.../ENG-993`), the underline runs from the 3rd character to two cells past the end, and the URL is not clickable on its visible text. This is how I first noticed it — a bare, single-line, non-OSC-8 URL that simply would not click, while the identical URL written as an OSC 8 hyperlink (`[ENG-993](...)`) elsewhere in the same buffer worked fine (OSC 8 needs no detection, so it's unaffected).

### Likely cause

The hint → screen-column mapping appears to use character/codepoint offsets rather than display-cell widths, so glyphs with East-Asian-Width W/F (and emoji) before the URL desync the highlight from the rendered text.

### Environment

- rioterm 0.4.5
- Linux 6.18.33 (Arch), Wayland session, GNOME
- Default Rio font (no `[fonts]` config)

nikicat:
## Summary

Fixes #1619. The clickable-URL underline and hit-region were horizontally
misaligned when a wide (emoji/CJK) or multibyte glyph preceded the URL on the
same line.

## Details

Regex hint extraction pushed `cell.c()` for every column (spacer placeholders
included) and then used onig byte offsets directly as grid columns, so a wide
glyph or multibyte codepoint earlier on the line shifted the underline and click
target right by the difference between byte count and cell count. This threads a
byte-offset → grid-column map through both call sites and accounts for the
trailing wide-cell spacer.

## Testing

`cargo test -p rioterm` — adds regression coverage for emoji, CJK,
multibyte-narrow, post-processing, and the real-world `⏺ https://` trigger from
the bug report.

Closes #1619

### DIFF
--- a/frontends/rioterm/src/hints.rs
+++ b/frontends/rioterm/src/hints.rs
@@ -5,6 +5,45 @@ use rio_backend::event::EventListener;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 
+/// Extract the visible text of `line` together with a byte-offset → grid
+/// column mapping. Spacer cells (the trailing half of a wide glyph, and
+/// the `LeadingSpacer` placed at the soft-wrap boundary) are skipped:
+/// they carry a placeholder `' '` that is not a separate visible
+/// character and would otherwise desynchronize the byte-to-column map.
+///
+/// `byte_to_col[i]` is the grid column of the cell whose codepoint's
+/// UTF-8 encoding contains byte `i` of the returned string. Trailing
+/// whitespace is left intact so the mapping stays aligned across the
+/// full row.
+///
+/// Used by the regex hint pipeline to convert onig's byte offsets
+/// (which would otherwise mis-locate the click target when emoji or
+/// CJK glyphs precede a URL) back into grid columns.
+pub(crate) fn extract_line_text_with_cols<T: EventListener>(
+    term: &rio_backend::crosswords::Crosswords<T>,
+    line: Line,
+) -> (String, Vec<usize>) {
+    let grid = &term.grid;
+    let cols = grid.columns();
+    let mut text = String::with_capacity(cols);
+    let mut byte_to_col: Vec<usize> = Vec::with_capacity(cols);
+
+    for col in 0..cols {
+        let cell = &grid[line][Column(col)];
+        if cell.is_spacer() || cell.is_leading_spacer() {
+            continue;
+        }
+        let c = cell.c();
+        let len = c.len_utf8();
+        for _ in 0..len {
+            byte_to_col.push(col);
+        }
+        text.push(c);
+    }
+
+    (text, byte_to_col)
+}
+
 /// State for hint selection mode
 pub struct HintState {
     /// Currently active hint configuration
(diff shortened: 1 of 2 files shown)


nikicat:
Rebased onto current `main` (`d656326020`).

Worth flagging the overlap, since at a glance the bug now looks fixed: `1b9c0552af` ("fix hyperlint hints") independently fixed **half** of #1619. Changing `Column(start)` → `Column(line_text[..start].chars().count())` corrects multibyte-*narrow* glyphs, so the real-world `⏺ https://…` case from the issue report does align correctly on `main` today.

The wide-glyph half is still broken. `chars().count()` counts a CJK or emoji glyph as one column, but the grid stores it as two cells (glyph + trailing spacer), so the hit-box still drifts left by the number of preceding wide glyphs.

Substituting `main`'s current logic into this branch's test suite, five cases fail:

| line | `main` start col | correct |
|---|---|---|
| `😀 https://example.com/emoji` | 2 | 3 |
| `世界 https://example.com/cjk` | 3 | 5 |
| `🎉🎉🎉 https://example.com/party` | 4 | 7 |
| `a😀b⏺c https://example.com/mix` | 6 | 7 |
| `(😀 https://example.com/end)` | 3 | 4 |

while `⏺ https://linear.app/ENG-993` passes — precisely the half `1b9c0552af` addressed.

This PR threads a byte-offset → grid-column map (`extract_line_text_with_cols`) through both call sites, which covers both halves, and extends the end column onto the spacer cell when a match ends on a wide glyph so the click target spans the full visible glyph. Spacer and leading-spacer cells are skipped when building the map so the byte↔column correspondence stays aligned across the row.

CI is green and the branch is conflict-free against `main`.

PR #1629, opened 2026-05-30
feat: smart double-click selection (OSC 8 links + regex rules)

### LINKED ISSUE #1621: Feature Request: URL-aware double-click word selection (don't break on `:` in `://`)
## Summary

Double-clicking a URL like `https://cli.github.com/manual` selects only `//cli.github.com/manual` — the scheme (`https:`) is dropped because `:` is treated as a hard word boundary by semantic selection. Would be great if double-click selected the whole URL.

## Steps to reproduce

1. `echo "https://cli.github.com/manual"` in Rio
2. Double-click anywhere on the URL
3. Selection ends at the first `/` after the colon; `https:` is excluded

## Expected

The full `https://cli.github.com/manual` (and similarly for `http://`, `ftp://`, `git://`, etc.) is selected.

## Why simply removing `:` from the boundary set isn't sufficient

The current semantic-selection algorithm (in `rio-backend/src/crosswords/search.rs`, `inline_search_left`/`inline_search_right`) has a single, context-free predicate:

```rust
if !matches!(cell.square.wide(), Wide::Spacer | Wide::LeadingSpacer)
    && needles.contains(cell.square.c())
{
    return Ok(point);   // stop: this cell is a boundary
}
```

`needles` is the hard-coded `semantic_escape_chars` at `rio-backend/src/crosswords/mod.rs:484`:

```rust
let semantic_escape_chars = String::from(",│`|:\"' ()[]{}<>\t\0");
```

So the boundary check is purely \"is this single character in the set?\" with no lookahead and no notion of context. That means:

- **Leaving `:` in the set** (status quo) breaks URL selection — the reported issue.
- **Removing `:` from the set** fixes URLs, but regresses other useful boundaries. `file.rs:42`, `key:value`, `host:port`, `mm:ss` timestamps, `nginx.conf:13:5` diagnostics, `git@github.com:user/repo`, IPv6 addresses, etc. would all be selected as one giant word instead of their semantic pieces.

A correct fix needs **context-aware** boundary logic, e.g. one of:

1. **URL-scheme lookahead:** when the candidate boundary char is `:`, peek the next two cells on the same row — if they're both `/`, treat the `:` as part of the word (continue expanding). Covers the common URL case (`https://`, `http://`, `ftp://`, `git://`, `ssh://`, `file://`, …) without changing behavior for `file.rs:42` or `key:value`. Small, targeted change to `inline_search_left/right`.
2. **Configurable `semantic_escape_chars`:** expose the boundary set as a config option (as alacritty, kitty `select_by_word_characters`, and iTerm2 do). Lets users opt out of `:` themselves, but still has the over-selection downside above when they do.
3. **Both:** URL-aware default + a config knob.

The unused `_url_regex` at `crosswords/mod.rs:487` already enumerates the URL schemes the project considers \"URLs\" — its `://` schemes overlap exactly with what option (1) would cover, so the project already implicitly agrees on the URL shape.

## Environment

- Rio: built from `main` (commit 842b497789 on a local branch)
- OS: Linux (Wayland)

nikicat:
## Summary

Makes double-click selection URL- and token-aware so it stops truncating at
semantic boundary characters (`:` in particular). Implements #1621.

## Details

Two layers in front of the existing semantic selection:

1. **OSC 8 fast path** — double-clicking inside an OSC 8 hyperlink span selects
   the whole link across soft-wrapped rows. The span walk lives in
   `crosswords::hyperlink::hyperlink_span_at` and is now shared with the hint path
   so both agree on link boundaries.
2. **Smart-selection rules** — six precision-tagged regex rules ship as defaults
   (URL, `file:line:col`, UUID, IPv4, email, git SHA). The engine scans the
   soft-wrapped logical line containing the click and returns the highest-precision
   match covering it, falling back to the previous semantic behavior on no match.
   URL precision sits above IPv4 so `https://192.168.1.1/admin` selects whole.

A `[smart-selection]` config table lets users disable the engine, remove/override
built-in rules, or append their own (e.g. JIRA tickets) with hot reload. The
runtime rule list is the merge of defaults and user `[[smart-selection.rules]]`
layered by name (same name replaces, `enabled = false` removes, fresh name
appends); bad regexes log a warning and skip just that rule. `SmartSelector` owns
the compiled rules and the reload step, so a config change swaps the rule set
without rebuilding `Screen`. Documented in rio.5.

## Testing

`cargo test -p rio-backend` — rule merge/disable/override on reload, plus a
worst-case timing sentinel that fills an 8192-cell line and asserts one
`select_at()` stays well under budget.

Closes #1621

### DIFF
--- a/extra/man/rio.5.scd
+++ b/extra/man/rio.5.scd
@@ -403,6 +403,60 @@ This section documents the *[renderer]* table of the configuration file.
 
 	Default: _3_
 
+# SMART SELECTION
+
+This section documents the *[smart-selection]* table of the configuration file.
+
+Double-clicking inside the terminal selects the token under the cursor. The
+resolution order is:
+
+. OSC 8 hyperlink span (the producer-emitted link, including non-URL anchor
+	text and soft-wrapped spans).
+. Smart-selection rules listed here: each rule's regex is matched against the
+	soft-wrapped logical line containing the click; the highest-precision
+	match that covers the click wins, with longest match as the tiebreaker.
+. The classic semantic selection (word-by-word stopping at the separator
+	characters).
+
+Steps 2 and 3 only run when step 1 doesn't apply, so producers that emit
+OSC 8 always win.
+
+*enabled* = _true_ | _false_
+
+	Master switch for the rule engine. When _false_, double-click goes
+	straight from the OSC 8 fast path to the semantic fallback.
+
+	Default: _true_
+
+*rules* = [{ name = _"<string>"_, regex = _"<string>"_, precision = _<integer>_, enabled = _true_ | _false_ },]
+
+	User-defined rules layered on top of the built-in set. Matching by
+	*name* against a default _replaces_ that rule; setting *enabled = false*
+	on a default's name _removes_ it; a fresh name _appends_ a new rule.
+	*precision* is a _0..=255_ integer (higher wins, ties broken by match
+	length).
+
+	Built-in rules (with their precisions): *url* (200), *file_line* (150),
+	*uuid* (140), *ipv4* (130), *email* (130), *git_sha* (50). Patterns use
+	the regex-automata DFA subset — no look-around, ASCII word boundaries
+	only (_(?-u:\\b)_).
+
+	Example:
+
+```
+	[smart-selection]
+	enabled = true
+
+	[[smart-selection.rules]]
+	name = "jira-ticket"
+	regex = "[A-Z]{2,10}-\\\\d+"
+	precision = 100
+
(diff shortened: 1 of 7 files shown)


tauraamui:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

💀

nikicat:
Converting to draft to keep the review queue focused on small fixes. The feature is maintained and released in the [nikicat/rio fork](https://github.com/nikicat/rio) — if there's upstream interest in #1621, I'm happy to rebase and mark it ready.

Issue #1587, opened 2026-05-06
Unable render CJK Chinese characters


shiba2046:
The font looks strange, but not sure how to fix.

<img width="804" height="385" alt="Image" src="https://github.com/user-attachments/assets/72d01755-e915-4310-947c-655030ad97f3" />

```toml
[fonts]
size = 16

[fonts.regular]
family = "Consolas" # Matches a Nerd Font (Oh My Posh requirements)
# style = "normal"
weight = 400
style = "default"
```

Tried all sorts of fonts, it can't seem to be able to pick up.

Installed via winget in Windows 11.

raphamorim:
Hey there, thanks. which rio version is that one?

Yi-Qing:
I used version 0.4.2, see this change log, when 0.4.0 show as: no more font extras, rio will always find the glyph if the font is installed. 
When used version 0.3.x, we can config as:
```toml
extras = [
    { family = "思源黑体 CN" },
    { family = "微软雅黑" },
]
```

evolighting:
@Yi-Qing 
> I used version 0.4.2, see this change log, when 0.4.0 show as: no more font extras, rio will always find the glyph if the font is installed. When used version 0.3.x, we can config as:
> 
> extras = [
>     { family = "思源黑体 CN" },
>     { family = "微软雅黑" },
> ]

尝试大模型agent阅读源码 ( 使用的是opencode go, 配合 vscode copilot ), 提问如何配置字体fallback, agent告诉我, 从代码来看, 目前可以改成使用symbol-map 来按 Unicode 范围映射到特定字体

我在配置文件中增加了对应配置:
```toml
[[fonts.symbol-map]]
start = "4E00"
end = "9FFF"
font-family = "HarmonyOS Sans SC"
```
测试看起来是可以生效的.

---

Using a coding agent to dig into some source code (OpenCode.Go via Copilot). I asked about font fallback settings, and be told that I could use a symbol-map for Unicode range mapping.

Updated my config with code abave, and it works.

Yi-Qing:
> 我在配置文件中增加了对应配置:
> 
> [[fonts.symbol-map]]
> start = "4E00"
> end = "9FFF"
> font-family = "HarmonyOS Sans SC"
> 
> 测试看起来是可以生效的.

Cool, work good, 好使

raphamorim:
Man it was supposed to work without need config. I will take a further look on this issue

s-cerevisiae:
[...]

Issue #1468, opened 2026-03-28
Box Drawing Characters Not Rendering Correctly


freepicheep:
Even with `font.use-drawable-chars` set true, the box glyphs aren't rendering correctly on rounded borders. This is using the IDE menu completion in Nushell.

<img width="1414" height="512" alt="Image" src="https://github.com/user-attachments/assets/cb48db4d-9dd6-4cf3-a8e5-24a303ce3687" />

For reference, see how Ghostty renders it.

<img width="1516" height="518" alt="Image" src="https://github.com/user-attachments/assets/48f63651-79d8-44e3-93f9-d1bd547d65fc" />

Applies to #1072

raphamorim:
Rounded corners U+256D-2570 are now drawn natively as arcs (3274eb056b, with `use-drawable-chars = true`). Please retest on 0.4.12.

PR #1630, opened 2026-05-30
fix: trigger font fallback discovery on Linux/Windows shape paths

nikicat:
## Summary

Codepoints absent from the bundled CascadiaCodeNF (e.g. U+23BF `⎿`, surfaced by
Claude Code's output) rendered as tofu on Linux/Windows because the grid- and
UI-text shape paths used `find_best_font_match`, which only walks
already-registered fonts and returns `(0, false)` on a miss.

## Details

macOS already used `resolve_font_for_char`, which falls back to cascade discovery
(fontconfig on Linux, font-kit on Windows) and registers the discovered font. The
macOS/non-macOS split was incidental refactoring, not a deliberate platform
difference — the width-measurement path in `sugarloaf::font_cache` already used
the cross-platform resolver. This unifies both shape call sites on
`resolve_font_for_char`, which also removes a window-vs-window race where one
route's `font_resolve` cache could lock in font_id 0.

## Testing

`cargo test -p rioterm` — adds a Linux regression test that probes a fresh
`FontLibrary` to confirm the host can cascade-discover U+23BF (probe-first to
avoid false-failing on minimal CI), then asserts the grid resolver returns a
non-zero font_id.

Refs #1468, #1587

### DIFF
--- a/frontends/rioterm/src/grid_emit.rs
+++ b/frontends/rioterm/src/grid_emit.rs
@@ -1211,7 +1211,7 @@ impl GridGlyphRasterizer {
     }
 
     #[inline]
-    fn resolve_font(
+    pub(crate) fn resolve_font(
         &mut self,
         ch: char,
         style_flags: u8,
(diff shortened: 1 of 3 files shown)


raphamorim:
Thank you, updated the changes and took you as co-author. It's in the main branch, closing it.

PR #1632, opened 2026-05-30
fix: refresh Taffy root size when scaled margin changes

### LINKED ISSUE #1495: on opening new tab the bottom of the terminal is hidden until window resize

nikicat:
## Summary

Fixes #1495 — on opening a new tab the bottom of the terminal is hidden until the
window is resized.

## Details

When a second tab is created and the tab bar transitions from hidden to visible,
`update_scaled_margin` updated only the stored margin — Taffy's root size still
reflected the old, larger available area. The subsequent layout pass then kept the
panel at its old height and the PTY was never reflowed, so the bottom rows ended up
clipped behind the freshly appeared tab bar. `update_scaled_margin` now refreshes
the Taffy root size so the next layout pass uses the correct available area.

## Testing

`cargo test -p rioterm` — adds a regression test asserting the panel height shrinks
by the new top margin after `update_scaled_margin`.

Closes #1495

### DIFF
--- a/frontends/rioterm/src/layout/compute_tests.rs
+++ b/frontends/rioterm/src/layout/compute_tests.rs
@@ -647,3 +647,73 @@ fn test_split_inside_resized_panel_preserves_proportions() {
         "Bottom (bottom half) should be ~400px tall, got {bottom_h}"
     );
 }
+
+// --- update_scaled_margin ------------------------------------------------
+
+fn margin_test_dimension() -> ContextDimension {
+    let dims = TextDimensions {
+        width: 8.0,
+        height: 16.0,
+        scale: 1.0,
+    };
+    ContextDimension::build(
+        800.0,
+        600.0,
+        dims,
+        cell_for(dims),
+        1.0,
+        14.0,
+        Margin::all(0.0),
+    )
+}
+
+/// Build a single-panel grid off sugarloaf (positions come from
+/// `calculate_positions`, not `apply_taffy_layout`).
+fn single_panel_grid() -> ContextGrid<rio_backend::event::VoidListener> {
+    use crate::context::create_mock_context;
+    use rio_backend::event::VoidListener;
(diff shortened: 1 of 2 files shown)


cmoron:
Independently hit this (as #1495) and diagnosed the same root cause: the terminal margin is tracked in two places that must stay in sync — the `scaled_margin` field (render offset) and the Taffy root-node size, from which the row count / PTY winsize is derived. `update_scaled_margin` updated only the field, so the grid kept its full-height row count and the bottom rows overflowed off-window until the next window resize (the one path that runs `try_update_size`). Reproduced and verified the fix behaviour on Linux (WSLg), including the reverse transition (closing back to a single tab).

Fixing at the setter is the right call — every `update_scaled_margin` caller stays correct — and the regression test is a nice touch. I've closed my duplicate (#1699) in favour of this one. 👍

nikicat:
@raphamorim gentle ping on this one — it's a small, self-contained layout fix with independent validation: @cmoron hit the same bug (filed as #1495), diagnosed the same root cause, verified the fix on Linux/WSLg including the reverse transition, and closed his duplicate PR #1699 in favor of this one. CI is green and it's conflict-free against main. Happy to adjust anything.

raphamorim:
This exact fix landed in #1723 (4f538d9d95, update_scaled_margin now refreshes the layout root) and #1495 is closed. Sorry it crossed with your PR, the diagnosis here pointed at the right place. Thanks!

Issue #1622, opened 2026-05-30
Window title on Wayland CSD drops non-Cantarell glyphs (emoji, CJK, etc.) — no font fallback


nikicat:
## Summary
On Wayland (tested on GNOME 50 / Mutter 50 / Cantarell), any codepoint not present in the single font loaded by Rio's client-side title bar — emoji, CJK, many symbols — is silently dropped (rendered as `.notdef` / invisible advance). Only Cantarell-covered Latin/diacritic ranges render.

Reproduce in any shell:

```sh
printf '\e]0;hello 😀 你好 ★\a'
```

Expected: `hello 😀 你好 ★` visible in the title bar.
Actual: `hello       ★` — emoji + CJK are blanks (advance preserved, no glyph).

## Environment
- Rio 0.4.5 / 0.4.6 (workspace)
- Wayland session, `XDG_CURRENT_DESKTOP=GNOME`
- GNOME Shell 50.2, Mutter 50.1
- System fonts: Adwaita Sans (UI), Cantarell present, Noto Color Emoji available via fontconfig fallback

## Root cause
GNOME/Mutter on Wayland forces client-side decorations, so Rio draws its own title bar via the bundled `sctk-adwaita 0.10.1` (`ab_glyph` backend, enabled by the default `wayland-csd-adwaita` feature in `rio-window/Cargo.toml:29`).

The ab_glyph title renderer ([`sctk-adwaita-0.10.1/src/title/ab_glyph_renderer.rs:128-148`](https://github.com/PolyMeilex/sctk-adwaita/blob/v0.10.1/src/title/ab_glyph_renderer.rs#L128-L148)) loads **one** font (whatever `fc-match` returns for the GNOME `titlebar-font` preference, defaulting to Cantarell) and lays out every char with `font.scaled_glyph(c)`. There is **no fontconfig fallback chain** for unmapped codepoints — `outline_glyph` returns `None`, nothing is drawn, but `h_advance` is still consumed, so you get a sequence of blank slots.

Cantarell has no emoji and no CJK coverage:

```
$ fc-match -s 'Cantarell' :charset=1F600
(no match)
```

## Where this matters in Rio
- `rio-window/Cargo.toml:29` — `default = [..., \"wayland-csd-adwaita\"]`
- `rio-window/Cargo.toml:30-31` — `wayland-csd-adwaita-crossfont` (alt backend, also single-font) is commented out; `wayland-csd-adwaita-notitle` would just hide the title.
[...]

PR #1634, opened 2026-05-30
deps: move rio-window to smithay-client-toolkit 0.20 and sctk-adwaita 0.11.1

nikicat:
Rebased onto current `main` (`d656326020`). **This PR no longer uses a git dependency** — see below.

## What this does

- `smithay-client-toolkit` 0.19.2 → 0.20.0
- `calloop` 0.13.0 → 0.14.0 (required by sctk 0.20)
- `sctk-adwaita` 0.10.1 → 0.11.1
- `wayland-csd-adwaita` feature: `sctk-adwaita/ab_glyph` → `sctk-adwaita/cosmic-text`

Two files: `rio-window/Cargo.toml` (+13/-1) and `Cargo.lock`.

## Why 0.11.1 specifically

`sctk-adwaita`'s default `ab_glyph` title renderer draws from a single font face with no
shaping and no fallback, so any codepoint outside that face renders as a `.notdef` box in the
CSD title bar. Its `cosmic-text` renderer does shaping plus font fallback.

That renderer landed in PolyMeilex/sctk-adwaita#98 and **first shipped in 0.11.1**
(released 2026-07-22). 0.11.1 depends on `smithay-client-toolkit ^0.20`, which is why the
sctk bump comes along with it. 0.12.0 also has it but requires sctk 0.21, a larger migration.

An earlier revision of this PR carried a `[patch.crates-io]` git pin to sctk-adwaita master
plus a matching cargo vendoring hash in `pkgRio.nix`, because the renderer was merged but
unreleased. **0.11.1 made both unnecessary** — the pin and the nix hash are gone, and the
lockfile now contains no `source = "git..."` entries at all.

## Result

Window title set to `😀 世界 ✓ café Ω`, captured on a headless Mutter virtual monitor
(Mutter implements no `zxdg_decoration_manager_v1`, so the client draws its own decorations —
same path as a GNOME session):

| build | title bar renders |
|---|---|
| `sctk-adwaita` 0.10.1 + `ab_glyph` | `CSDBEFORE ▯ ▯▯ ✓ café Ω` — emoji and CJK are `.notdef` boxes |
| `sctk-adwaita` 0.11.1 + `cosmic-text` | `CSDAFTER 😀 世界 ✓ café Ω` — all glyphs resolve |

Addresses #1622.

## Note on scope

`wayland-csd-adwaita` is in `rio-window`'s own `default` feature list, but the workspace
declares `rio-window = { …, default-features = false }` and `rioterm`'s `wayland` feature
enables only `rio-window/wayland` and `wayland-dlopen`. So `cargo tree -p rioterm -i
sctk-adwaita` currently resolves to nothing, and I had to pass
`--features rio-window/wayland-csd-adwaita` explicitly to produce the comparison above.

In other words: the dependency bumps here affect every Linux build, but the
`ab_glyph → cosmic-text` switch only takes effect for builds that opt into
`wayland-csd-adwaita`. Whether `rioterm` should enable it by default is a separate question
with a real dependency-weight tradeoff, so I've left it out of this PR rather than bundling it.

No `rio-window` source changes are needed for the sctk bump: keyboard input goes through a raw
`Dispatch<WlKeyboard>` impl, so the `KeyboardHandler::update_modifiers` signature change in
sctk 0.20 does not apply, and no calloop 0.14 breaking API is used.

CSD title bar: ab_glyph renders emoji and CJK as .notdef boxes; cosmic-text renders them correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,22 +3,6 @@
 version = 4
 
 [[package]]
-name = "ab_glyph"
-version = "0.2.32"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "01c0457472c38ea5bd1c3b5ada5e368271cb550be7a4ca4a0b4634e9913f6cc2"
-dependencies = [
- "ab_glyph_rasterizer",
- "owned_ttf_parser",
-]
-
-[[package]]
-name = "ab_glyph_rasterizer"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "366ffbaa4442f4684d91e2cd7c5ea7c4ed8add41959a31447066e279e432b618"
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -961,6 +945,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "core_maths"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77745e017f5edba1a9c1d854f6f3a52dac8a12dd5af5d2f54aecf61e43d80d30"
+dependencies = [
+ "libm",
+]
+
+[[package]]
 name = "coreaudio-rs"
 version = "0.11.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


nikicat:
Companion PR opened: #1768 enables `wayland-csd-adwaita` in rioterm's `wayland` feature, which is what makes this one take effect for the rio binary. They're independent — either can merge first — and neither closes #1622 alone. #1768 has the dependency-cost breakdown for both states.

Issue #1015, opened 2025-03-17
Not rendering 𝓐


ziraqyoung:
Rio terminal isn't rendering the **𝓐** glyph (Mathematical Script Capital A, Unicode `U+1D4D0`)

```bash
echo -e "\U1D4D0"
```

I have tried a couple of fonts: FiraCode Nerd Font, JetbrainsMono Nerd Font (both are rendering empty box).

The default rio font Cascadia code also doesn't render.

![Image](https://github.com/user-attachments/assets/499d53e9-f240-44cf-8f81-7887f16f8bfe)

I am not sure if this is related to my system or not rendering in rio.

pinpox:
Can reproduce this. It also fails to render other characters.

raphamorim:
Should be fixed in main

pinpox:
Can confirm it works on main:

<img width="857" height="1084" alt="Image" src="https://github.com/user-attachments/assets/6f068f27-98a0-41e9-bf30-6c7de5183350" />

- Bottom is current relase version (the icons for the folders are broken)
- Top is main

The colors are a bit different (e.g. the black background color), is that expected? did the config format change?

raphamorim:
> The colors are a bit different (e.g. the black background color), is that expected? did the config format change?

I didn’t change but wonder if is wgpu related updates

pinpox:
Sorry my mistake. I was still running the config from this PR https://github.com/raphamorim/rio/pull/1358 and main does not pick up the theme correctly yet. I rebased the PR onto newest main and now everything works and looks as expected. I'll keep my fork until that is merged, as the feature is essential to me.

PR #1642, opened 2026-06-03
fix: restore font fallback discovery on Linux/Windows

pinpox:
The terminal render path (rioterm grid_emit::resolve_font) and the UI text path (sugarloaf text) resolved each glyph's font through find_best_font_match, which only consults already-loaded fonts and returns the primary font (font_id 0) on a miss, it never triggers cascade discovery. So on Linux/Windows any codepoint absent from the user's primary font (Nerd Font icons from `ls --icons`, CJK, ...) rendered as the primary font's .notdef tofu.


ziraqyoung:
Looks like this PR fixes an issue I had opened, https://github.com/raphamorim/rio/issues/1015.

raphamorim:
Thank you, updated the changes and took you as co-author. It's in the main branch, closing it.

PR #1645, opened 2026-06-04
fix: apply kitty X=/Y= sub-cell pixel offset to placements

floens:
I was trying out Rio for its kitty image support (Rio is the only term emulator that works on Windows AND has good KIP support!) and I noticed that the sub-cell pixel offsets were not wired in yet.

Was this by accident?

Easy to wire in, connecting it from the parser (that already parses it) to the renderer. I tested this locally.


raphamorim:
@floens thanks for the kind words, i did few updates to your branch. could you check [#pr-145](https://github.com/raphamorim/rio/pull/1646)

floens:
Thanks! I'm not comfortable in this codebase at all. Thanks for picking this up, will close this PR in favor of #1646.

PR #1650, opened 2026-06-07
support width on glyph protocol but only renderer level

raphamorim:
This branch is completely unpolished but here's the idea:

A Glyph Protocol registration does NOT change a codepoint's logical width: the cell layout stays consistent with system `wcwidth` so the cursor / selection / copy never desync with a width-unaware consumer (a shell line editor, etc.). The declared `width` is honoured purely at render time, where the glyph overflows into the following cell(s) in pixels.

It counters #1649 . So if this merges we don't need to drop width from the spec.

Due to the nature of the terminals, we will only have width respected in the render level.

image

However you may known that if width is 2, then applications should also give a space themselves.

and ofc you can renders as well with width 1:

image


raphamorim:
Reviewed and updated this branch against current main:

- Restored main's pty read machinery: the original commit accidentally carried a revert of the buffered read (READ_BUFFER_SIZE back to 4096 and MAX_LOCKED_READ deleted) and both merges kept it. Unrelated to this PR and a real perf regression if merged; `rio-backend/src/performer/mod.rs` is now byte-identical to main.
- Fixed the doc comments that still described the abandoned layout-authoritative design ("authoritative for all terminal layout decisions", "spec 8.5") on `GlyphCommand::Register`, the `Handler::glyph_register` trait doc and `RegisteredGlyph::width`. They now match what the code and its test actually do: width is a render span, layout stays at wcwidth.
- Removed `GlyphRegistry::width_of`, which had zero callers and a doc claiming an input-hot-path use that intentionally does not exist in this design.
- Small cleanups: stray blank line, rustfmt, and the `rasterize_payload` doc now mentions the caller passes the span box rather than one cell.
- Verified visually with a dev build: a width=2 registration paints across two cells while the cursor advances one cell (marker lands per wcwidth); width=1 stays contained; overwriting the overflow cell shares pixels as documented; register replies come back with status=0. Full suites green (622 backend, 153 sugarloaf, 168 rioterm).
- The site's glyph-protocol page still documented width as layout-authoritative (pre-#1649 wording); updated to render-span semantics.

What held up well under review: the `glyph_cell_scale` contain-guard plus MAX_RASTER_DIM makes hostile `upm` values harmless, and moving bearings out of the cached slot into emit time keeps the atlas entry a pure (codepoint, version, scale) mapping.

PR #1667, opened 2026-06-19
fix: honor surface alpha in Vulkan (restore Wayland background transparency)

agjini:
Since the native Vulkan backend became the default on Linux (in v0.4.0), `window.opacity` and transparent backgrounds stopped working on Wayland. The swapchain was created with `CompositeAlphaFlagsKHR::OPAQUE` hard-coded, which tells the compositor to ignore the surface's alpha channel entirely.

This selects the composite-alpha mode from `surface_capabilities.supported_composite_alpha`, mirroring the wgpu backend's preference order (post-multiplied → pre-multiplied → inherit → opaque). On Wayland this resolves to `PRE_MULTIPLIED`, matching the pre-Vulkan (wgpu) behavior from v0.3.9.

Fixes the transparency regression between v0.3.9 and v0.4.0.


agjini:
@raphamorim can you have a look, the fix seems to be simple. Thanks

raphamorim:
Sorry late reply, the changes were looking good. Just wanted to try it out. Thank you!

PR #1671, opened 2026-06-20
Fix macOS font fallback for bold/italic family resolution

guibeira:
## Summary
wrong font rendered because of fallback
image
After the fix:
image

  Fix macOS font resolution for families where CoreText returns no candidates for `family + bold/italic trait`.

  When the trait-based lookup fails, Rio now falls back to a family-only lookup and uses the existing scoring logic to pick the best matching face, such as Bold, Italic, or Bold Italic.

  ## Why

  Some installed font families, like `UbuntuMono Nerd Font`, expose their Bold face correctly, but CoreText returns zero candidates when queried with `family + bold trait`.

  This caused bold text to fall through to a different fallback font, which could produce mismatched glyph metrics and visible vertical misalignment for punctuation.

### DIFF
--- a/sugarloaf/src/font/macos.rs
+++ b/sugarloaf/src/font/macos.rs
@@ -589,12 +589,28 @@ pub fn find_font_path(
             unsafe { CFString::wrap_under_get_rule(kCTFontStyleNameAttribute) };
         pairs.push((style_key, CFString::new(name).as_CFType()));
     }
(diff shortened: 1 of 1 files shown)


raphamorim:
Merged in main with few updates, thank you!

guibeira:
vlw Rapha!

Issue #1668, opened 2026-06-19
bug: hint labels not showing up on windows


CovetingEpiphany2152:
rioterm 0.4.7 on Windows 10.

With the following toml, after triggering the keybindings, the matched text are highlighted, but they do not have any label. So the user wouldn't know which button to press to initiate the hint command or actions.

```toml
[[hints.rules]]
regex = "(ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:)[^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+"
hyperlinks = true
post-processing = true
persist = false

[hints.rules.command]
command = { program = "C:\\Program Files\\Mozilla Firefox\\firefox.exe", args = ["--new-tab"] }

[hints.rules.mouse]
enabled = true
mods = ["Alt"]

[hints.rules.binding]
key = "."
mods = ["Control", "Shift"]
```

<img width="901" height="115" alt="Image" src="https://github.com/user-attachments/assets/88f9e23e-2a02-46ed-a973-92dd31927cfb" />

CovetingEpiphany2152:
same issue for the following:

```toml
[hints]
alphabet = "jfkdls;ahgurieownvcm,x"

[[hints.rules]]
regex = "/?(?:[\\w.-]+/)*[\\w.-]+"
hyperlinks = false
post-processing = true
persist = false

[hints.rules.action]
action = "Copy"

[hints.rules.mouse]
enabled = true
mods = ["Alt"]

[hints.rules.binding]
key = "/"
mods = ["Control", "Shift"]
```

OrionOth:
Same issue with Linux, Vulkan, 0.4.7, no/default hints configuration.

raphamorim:
Hint label rendering was rebuilt in 1b9c0552af (0.4.9+): labels now draw as overlays with corrected column offsets. Please retest on 0.4.11 and reopen if labels are still missing on Windows.

PR #1674, opened 2026-06-23
fix: render hint labels and correct match column offsets

bigboss2063:
Issue #1668: hint labels not showing up
- Wire hint_labels through PanelFrame into build_row_bg (paints hint_background on labeled cells) and build_row_fg (new emit_hint_label_glyphs phase shapes each label char as a one-cell run and overlays it in hint_foreground, drawn last so it replaces the underlying matched text). The configured hint_foreground / hint_background colors were defined but read by no renderer.

Byte-offset-as-column drift on lines with non-ASCII cells
- find_regex_matches and find_regex_match_at_point converted oniguruma byte offsets to column indices via char count. The previous Column(byte) drifted matches rightward on lines containing non-ASCII cells (e.g. nerd-font PUA icons from oh-my-posh prompts), breaking highlight/label placement and mouse-hover hit-testing. Copy/Command actions still worked because they use the text field, not the Pos.

Stale hint_labels during search
- update_hint_state now clears hint_labels whenever hint mode is inactive (not only when search is also inactive); previously leftover labels could be drawn over search results once the label renderer was wired in.

Fix result (using the same configuration as in the issue)

Before：
image

After：
image
image

### DIFF
--- a/frontends/rioterm/src/grid_emit.rs
+++ b/frontends/rioterm/src/grid_emit.rs
@@ -1006,18 +1006,26 @@ pub fn build_row_bg(
     term_colors: &TermColors,
     row_sel: Option<RowSelection>,
     row_hints: &[RowHint],
+    // Columns on this row that carry a hint label glyph. These cells
+    // get `hint_background` painted over the match highlight so the
+    // label letter (overlaid in `build_row_fg`) stays readable. Sorted
+    // is not required — scanned linearly, matches the small-N case
+    // (hint labels are bounded by visible matches).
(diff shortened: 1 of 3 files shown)


raphamorim:
Both halves of this landed via 1b9c0552af: hint label rendering (push_hint_label_styles / overlay_hint_labels, with tests) and the byte-to-char column fix in hints.rs. Closing as superseded; #1668 stays open until labels are confirmed working on Windows. Thanks for the fix!

PR #1675, opened 2026-06-24
[Bug] Fix macOS titlebar behavior when the tab island is hidden for a single tab.

Ayasser1:
## Summary

  Fix macOS titlebar behavior when the tab island is hidden for a single tab.

  When Rio had only one tab, the hidden island/titlebar area still behaved like terminal text for some mouse interactions. This made the cursor show incorrectly over the traffic-light/titlebar area and prevented
  double-click maximize/restore from working there.

  This updates the macOS top chrome handling so:

  - the cursor stays as the arrow over the titlebar/traffic-light band
  - double-clicking the hidden single-tab top chrome toggles maximize/restore
  - single-click dragging still falls through to the existing manual window drag path

  ## Testing

  ```sh
  cargo fmt -- --check
  cargo check -p rioterm --features wgpu
  ```

 ## Manual test:

  - launched Rio with one tab
  - hovered over the macOS titlebar/traffic-light area
  - confirmed the cursor stays as an arrow
  - double-clicked the top chrome and confirmed the window maximizes/restores
  
### Before

https://github.com/user-attachments/assets/46aa560a-f362-4322-8599-109ecf4b7573

### After

https://github.com/user-attachments/assets/d520a0a0-8c23-4255-8739-2b74e144dc41

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1411,17 +1411,17 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     return;
                 }
 
-                // Only force the default cursor while the island is
-                // visible — when it's hidden (hide_if_single + single
-                // tab on macOS) the band at the top has no tabs to
-                // hover, and the I-beam from the terminal grid below
-                // should stay during top-edge drags.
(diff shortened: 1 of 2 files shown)


Ayasser1:
@raphamorim 👀

PR #1678, opened 2026-06-25
fix: hot-reload key bindings on config update

cantona:
`update_config` re-applied fonts, colors, cursor and keyboard config but never rebuilt the key bindings, so changes to the `[bindings]` section required a full restart. Rebuild bindings from the new config so binding edits reload live like the rest of the config.


raphamorim:
This landed independently in #1717: bindings are now rebuilt on config reload (frontends/rioterm/src/screen/mod.rs). Thanks for the fix, closing as superseded.

Issue #1620, opened 2026-05-29
[Bug] Wayland: mouse selection is not exported to the PRIMARY selection (middle-click / Shift+Insert never paste Rio's own selection)


nikicat:
### Summary

On Wayland, selecting text with the mouse in Rio does **not** claim the PRIMARY selection. As a result, middle-click paste and `Shift+Insert` paste whatever *other* application last owned the primary selection — never the text just selected in Rio. Every other Wayland terminal I have (foot, Alacritty, kitty, GNOME Console) claims PRIMARY on selection, so this is surprising.

### Steps to reproduce

Using `wl-clipboard` to observe the selections:

```sh
# 1. Seed PRIMARY from another source
wl-copy --primary "SENTINEL_BEFORE_RIO_SELECT"
wl-paste --primary            # => SENTINEL_BEFORE_RIO_SELECT  (as expected)

# 2. Now select some text with the mouse inside the Rio window, then:
wl-paste --primary            # => STILL "SENTINEL_BEFORE_RIO_SELECT"
                              #    expected: the text just selected in Rio
```

Equivalently: select text in Rio, then middle-click (or `Shift+Insert`) in any other window — you get the stale primary contents, not Rio's selection.

### Actual behavior

After selecting text in Rio, the Wayland PRIMARY selection is unchanged. (The clipboard is also unchanged, but that is the separate `copy-on-select` path — see below.)

### Expected behavior

Selecting text should claim the Wayland primary selection (via `zwp_primary_selection_device_manager_v1`), so middle-click / `Shift+Insert` paste the current Rio selection — matching other terminals and the standard X11/Wayland selection model.

### Notes

- This is independent of `copy-on-select`. `copy-on-select = true` mirrors the selection into the **clipboard**; this report is about the conventional **primary-on-select** behavior, which should happen regardless. (Related but distinct: #1487, #1536, #1494, all clipboard-side.)
- Paste itself works: `Shift+Insert` already pastes PRIMARY fine — the bug is purely that Rio never *writes* to PRIMARY.

### Environment

- rioterm 0.4.5
- Linux 6.18.33 (Arch), Wayland session, GNOME/Mutter
- Default Rio font

ildar:
I see at least two PR fixing this. ETA ?

ildar:
JFYI I built Rio with https://github.com/raphamorim/rio/pull/1679 and enjoy it "happily ever after".

yuelwish:
[...]

PR #1679, opened 2026-06-25
fix: copy selection to primary buffer for middle-click paste

cantona:
On mouse drag-release the selection was only copied to the clipboard (and only when `copy-on-select` was enabled), never to the primary selection buffer, so middle-click paste had nothing to paste. Always mirror the selection into the primary buffer on release, matching standard X11/Wayland behavior. `copy_selection` is a no-op on an empty selection.


ildar:
ref: https://github.com/raphamorim/rio/issues/1620

PR #1681, opened 2026-06-25
feat: configurable tab title font size via navigation.tab-font-size

cantona:
The tab title font size was hardcoded (`TITLE_FONT_SIZE`). Add a `navigation.tab-font-size` config option so the tab strip text size can be tuned.


cantona:

PR #1682, opened 2026-06-25
fix: pixel-snap tab-bar border lines so they don't randomly disappear

cantona:
The vertical tab separators and the inactive-tab / left-margin bottom borders were drawn 0.5 logical px thick at fractional positions. `sugarloaf.rect` scales logical coords by `scale_factor` without rounding, so these thin lines landed between physical pixels and the rasterizer dropped them — a window resize shifted positions enough to occasionally realign them (the classic 'separator missing, reappears on resize'). Snap each line's position to a whole physical pixel and use a 1-physical-px thickness so they always render.

Note: this branch is stacked on the tab-font-size (#1681) and tab-bar-height work, so it includes those commits; the border-fix commit is the last one. Happy to rebase onto main once those land.


cantona:

PR #1683, opened 2026-06-25
feat: configurable tab bar height via navigation.tab-bar-height

cantona:
Adds a `navigation.tab-bar-height` config option (default 34) to set the height of the tab/navigation bar (the 'island'), in logical pixels. Previously the height was a hardcoded `ISLAND_HEIGHT` const. The value is plumbed through to the Island renderer and the top-bar hit-test sites, and hot-reloads on config change. Mirrors the existing `navigation.tab-font-size` option.

Note: stacked on the tab-font-size work (#1681), so it includes that commit; the tab-bar-height commit is the last one. Happy to rebase onto main once #1681 lands.


cantona:

PR #1685, opened 2026-06-25
fix: keep quit-confirmation dialog visible on idle frames

cantona:
The ConfirmQuit dialog is drawn into sugarloaf before `screen.render()`, but `render()` calls `discard_frame()` whenever `should_present` is false — i.e. no panel damage and no animation, which is exactly the case while the dialog is up over an idle terminal. The dialog draws were discarded and only reached the screen on frames a cursor blink (or other animation) happened to force, so the prompt flickered in and out ('sometimes I can't see it').

Add a `force_present` flag to `render()`, set while the route is `ConfirmQuit`, so the dialog frame is always presented.

Standalone, based on main.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1842,7 +1842,10 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                             );
                         }
 
-                        if let Some(window_update) = route.window.screen.render() {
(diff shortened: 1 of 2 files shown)


cantona:
no needed. 89c5744eff8e6a21e1e86bf523265a03c63ad018 fixed it

PR #1686, opened 2026-06-25
fix: dismiss quit confirmation dialog on a single n/Esc press

cantona:
Pressing **n** (or Esc) in the quit-confirmation dialog required two presses to actually cancel, while **y** worked on the first press.

Cause: the `n`/`Escape` branches set the route back to `Terminal` but never requested a redraw. On an idle terminal the next frame is discarded (`should_present == false`), so the dialog stayed painted until some later event forced a repaint — making the first press look ignored. `y` was unaffected because `quit()` tears the window down.

Fix: call `request_overlay_redraw()` on dismiss (matching the assistant/command-palette dismiss paths) so the dialog clears on the first press.

Standalone, based on main.


cantona:
no needed. 89c5744eff8e6a21e1e86bf523265a03c63ad018 fixed it

PR #1687, opened 2026-06-25
fix: recompute grid rows when the tab bar appears/disappears

cantona:
When `hide_if_single` is on, creating a second tab makes the tab bar appear, which grows `scaled_margin.top` to reserve its band. But `update_dimensions` re-laid out via `apply_taffy_layout` **without refreshing Taffy's available height** for the new margin — so the grid kept its previous (taller) row count and the bottom row was clipped off the bottom of the window. It only corrected itself after a window resize, because `resize()` calls `try_update_size` (which does push the new margin into Taffy).

Fix: `update_dimensions` now calls `try_update_size(self.width, self.height)` before `apply_taffy_layout`, so the available height accounts for the tab-bar band and the row count is recomputed. Also fixes the reverse transition (closing back to one tab).

Standalone, based on main.


raphamorim:
Superseded by #1723: update_scaled_margin now re-runs try_update_size whenever margins change, which covers the tab strip toggling. #1495/#1528 are closed. Thanks for the diagnosis.

Issue #1661, opened 2026-06-13
Macos - Tab not closed after exit


duongtugiang:
Clear Reproduction Steps
I just updated to the newest Rio version and found a regression. Here is the exact behavior:
1. Launch Rio and open multiple tabs (e.g., 3 tabs).
2. In two of those tabs, type exit and press Enter.
3. The affected tabs freeze for a few seconds and do not close.
4. After waiting a few seconds, click inside one of the frozen tabs — it immediately closes/disappears.
---
Version Behavior
• Last working version: 0.3.11
• Bug appears starting from: any version after 0.3.11
Mac Sequoia

cantona:
Fix: https://github.com/raphamorim/rio/pull/1688

PR #1688, opened 2026-06-25
fix: repaint immediately when a tab closes on shell exit

cantona:
With several tabs open, typing `exit` in a tab sometimes closed it instantly and sometimes left it frozen on screen (showing the dead shell) until you clicked another tab.

On shell exit the PTY thread sends `CloseTerminal(route_id)` **and** a separate `Render`. The `CloseTerminal` handler closed the tab and resized the strip but never requested a redraw itself — it relied on that separate `Render`. If the `Render` was processed before `CloseTerminal` (or coalesced away), the post-close state was never painted, so the closed tab lingered until a later event forced a repaint.

Fix: `mark_dirty()` + `request_redraw()` in the `CloseTerminal` handler so the close is shown immediately, regardless of event ordering.

Standalone, based on main.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -512,6 +512,14 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     } else {
                         let size = route.window.screen.context_manager.len();
                         route.window.screen.resize_top_or_bottom_line(size);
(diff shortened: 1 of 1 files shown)


cantona:
Fix issue https://github.com/raphamorim/rio/issues/1661

PR #1689, opened 2026-06-25
fix: parse function-key and enter bindings in config [bindings]

cantona:
Function-key bindings in the `[bindings]` config silently did nothing. `convert()` (the config key parser) had no `f1`-`f12`, no `enter` alias (only `return`), and no `backspace`, so those keys hit the catch-all `Err` arm and the binding was dropped. For example:

```toml
[bindings]
keys = [{ key = "f11", action = "togglefullscreen" }]
```

did nothing — F11 fell through to the terminal as its escape sequence.

Add `f1`-`f12`, an `enter` alias, and `backspace` to `convert()`, matching the key set the hint-binding parser already accepts.

Standalone, based on main.


raphamorim:
Superseded by #1717, which added backspace, return/enter and f1-f12 to the binding key parser (bindings/mod.rs, with tests). Thanks!

PR #1690, opened 2026-06-25
feat: add bindable togglemaximized action

cantona:
`Action::ToggleMaximized` existed but was never wired up — it had no action-string mapping (so `[bindings]` couldn't reference it) and no handler. This adds:
- `"togglemaximized"` in the action-string parser
- `toggle_maximize_window()` emitting `RioEvent::ToggleMaximized`
- the `Act::ToggleMaximized` arm
- an application handler that flips `winit_window.set_maximized()` (works on Wayland)

So users can bind window maximize/restore, e.g. `{ key = "m", with = "super", action = "togglemaximized" }`.

Standalone, based on main.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1011,6 +1011,12 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


cantona:
@raphamorim this is quality of life little patch. Please comment. Thanks.

PR #1694, opened 2026-06-29
feat: terminal triggers (iTerm2-style regex → action rules)

cantona:
## Terminal triggers

Adds iTerm2-style **triggers**: user-defined regex rules matched against
terminal output that fire an action. Rules live in a dedicated
`~/.config/rio/triggers.toml` (loaded independently of `config.toml`, so
a triggers-only setup needs no `config.toml`) — mirroring how themes load
from a separate file. The file's top level is the rule list (`[[rules]]`).

### Actions (`snake_case` in TOML)
- **notify** — desktop notification; optional `urgency = "low" | "normal" | "critical"`
- **tab_color** — recolor the tab that matched
- **run** — spawn a detached command (output discarded)
- **send_text** — type text into the matched session's PTY
- **coprocess** — run a command and write its stdout back into the PTY
- **highlight** — colorize the matched text in the grid

`run` / `send_text` / `coprocess` / `notify` support `\0..\9`
capture-group substitution.

### `feed_screen` (coprocess option)

By default a coprocess only receives the regex capture groups as args.
Set `feed_screen = true` to also pipe the **visible screen plus up to
200 lines of scrollback above it** to the command's **stdin** — so the
command can act on multi-line context (e.g. a wrapped block, a whole
error trace) that a single per-line capture can't carry. The command
reads that on stdin; whatever it writes to stdout is typed back into the
PTY as usual.

Details: the payload is captured lazily (only on the first `feed_screen`
match in a frame — non-matching frames pay nothing) and byte-capped
comfortably under the OS pipe buffer (~64KB on Linux), truncated from
the top so the newest lines / visible prompt are always kept, avoiding a
stdin/stdout deadlock.

```toml
# Send the whole visible block to a summarizer, type its reply back
[[rules]]
regex = "^SUMMARIZE$"
action = { coprocess = { program = "my-summarizer", feed_screen = true } }
```

### Matching
- Scans the **focused pane's** output as it renders; a background tab is
  not scanned until focused (so a notify fires when you switch to it, not
  when the text first appeared).
- **Damage-gated**: only lines the terminal marked dirty are re-scanned,
  keyed by a scroll epoch so a line's identity is stable across scrollback
  movement. Each rule fires once per fresh occurrence, not every frame —
  prompt redraws and alternate-screen TUIs still register new output, but
  a static match on screen isn't re-fired.
- Per-rule **`instant`** fires on the live cursor line (prompts with no
  trailing newline).
- Per-rule **`once`** fires at most once until re-armed; the bindable
  **`resettriggers`** action re-arms them (keeping the per-line dedup so
  stale on-screen content isn't re-fired). Together these enable
  multi-step automations.

### Robustness
- A malformed `triggers.toml` surfaces the parse error in the UI and keeps
  the previously-loaded rules live instead of silently dropping them.
- `coprocess` has a bounded deadline so a hung command can't wedge the
  scan.

### Design
- `onig` (Oniguruma) regexes — the same engine `[hints]` already uses.
  `onig::Regex` is `!Send`, so matching runs on the main thread against
  the terminal already locked for rendering; the backend stays
  `Send`-safe (config structs + a `TriggerFired` event payload only).
- Per-line text extraction is factored out of `hints.rs` into a shared
  `extract_line_text` helper (no duplication).
- Zero cost when unused: with no `triggers.toml` / no rules, nothing is
  locked or scanned.

### Example `triggers.toml`
```toml
# Turn the tab red on any "error"
[[rules]]
regex = "(?i)\\berror\\b"
action = { tab_color = { color = "#FF0000" } }

# Desktop notification with a capture group
[[rules]]
regex = "error: (.*)"
action = { notify = { title = "Error", body = "\\1", urgency = "critical" } }

# Highlight WARN mid-line (before Enter)
[[rules]]
regex = "WARN"
instant = true
action = { highlight = { color = "#FFAA00" } }

# Run a command and type its output back
[[rules]]
regex = "^run date"
action = { coprocess = { program = "date" } }
```

### Notes
- **Security**: `run` / `coprocess` (spawn processes) and `send_text`
  (types into the shell) act on text that may be attacker-controlled
  terminal output — the same command-injection surface iTerm2 documents.
  No shell is involved and words are never re-split, but a capture can
  begin with `-`, so put `--` before capture arguments where the program
  supports it. Entirely opt-in: off until `triggers.toml` exists.

CI: `cargo fmt --check`, `cargo clippy`, and `cargo test` all green.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -59,6 +59,17 @@ impl Application<'_> {
         if let Some(error) = config_error {
             router.propagate_error_to_next_route(error.into());
         }
+        // A broken triggers.toml must be as visible as a broken
+        // config.toml; a config error and a triggers error can't both
+        // be pending here (a failed config load never loads triggers).
+        if let Some(message) = &config.triggers_load_error {
+            router.propagate_error_to_next_route(rio_backend::error::RioError {
+                report: rio_backend::error::RioErrorType::InvalidTriggersFormat(
+                    message.clone(),
+                ),
+                level: rio_backend::error::RioErrorLevel::Warning,
+            });
+        }
 
         let proxy = event_loop.create_proxy();
         let event_proxy = EventProxy::new(proxy.clone());
@@ -145,8 +156,8 @@ impl Application<'_> {
         }
     }
 
-    fn handle_desktop_notification(&self, title: &str, body: &str) {
-        rio_notifier::send_notification(title, body);
+    fn handle_desktop_notification(&self, title: &str, body: &str, urgency: u8) {
+        rio_notifier::send_notification(title, body, urgency);
     }
 
     pub fn run(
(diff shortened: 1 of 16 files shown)


cantona:
@raphamorim this is very useful. Please consider and comment.

PR #1695, opened 2026-06-30
fix: drop unused SurfaceCapabilitiesKHR import (sugarloaf vulkan)

cantona:
`SurfaceCapabilitiesKHR` is imported in `sugarloaf/src/context/vulkan.rs` but never used — left over from #1667 (the composite-alpha change). It emits an `unused_imports` warning on a `--no-default-features --features=wayland` build. This drops it; `CompositeAlphaFlagsKHR` (the one actually used) stays.


cantona:
Closing — the same fix landed upstream as part of the wgpu 30 upgrade.

PR #1702, opened 2026-07-06
fix: reporting of support for sixel and iterm2 graphics

jquast:
Since these have been unsupported since 0.4, it is good not to report support of them, so downstream applications do not use them until they are re-integrated, favor kitty when supporting.


cantona:
fixed at #1703  also i have a list of PR hope someone can review

raphamorim:
Closing since the premise no longer holds on main: sixel and iTerm2 rendering was restored in #1729/#1733 (ships in 0.4.12), so the DA support reporting should stay as is. It was the right call against 0.4.x at the time, thanks for the PR and for the detailed protocol notes.

PR #1428, opened 2026-01-18
Enable TIFF decode for iTerm2 images

zouyonghe:
## Summary
  - Enable TIFF decoding in `image` crate to support iTerm2 image protocol output from chafa/PixelTerm.

  ## Context
  Rio parses OSC 1337 and feeds the payload to `image::load_from_memory`.
  Chafa’s iTerm2 renderer emits TIFF data, which previously failed to decode.

  ## Changes
  - Add `tiff` feature to `image` dependency.

  ## Testing
  - manual verification: PixelTerm `--protocol iterm2` now displays images.

PR #1703, opened 2026-07-08
fix: render sixel and iTerm2 inline images (imgcat) again

### LINKED ISSUE #1591: iTerm2 image protocol broken on 0.4.2 & 0.4.3
Tried changing renderer to `webgpu` and `vulkan`, both same. I don't know if it's an arch-repo error again like someone said here before.  I confirm`icat`, `imgcat`, `viu` all 3 used to work before.

> rio is working fine when I compile myself @imgurbot12 but not when I try it from arch's repos. This seems like it may be a arch specific issue. Likely a good idea to report there. 
> 
> I am not sure what arch is doing wrong, but here is the build script for it https://gitlab.archlinux.org/archlinux/packaging/packages/rio/-/blob/main/PKGBUILD?ref_type=heads
> 
>  

 _Originally posted by @Quackdoc in [#999](https://github.com/raphamorim/rio/issues/999#issuecomment-2785067944)_

Image

Inside tmux (both `vulkan` & `webgpu`)

Image

```bash
> rio --version
rioterm 0.4.3

> fastfetch
 DISTRO 󰑃  CachyOS x86_64
│ ├ 󰑃  Linux 7.0.3-zen1-2-zen
│ ├󰏖 󰑃  1348 (pacman)
│ └ 󰑃  nvim
 DE/WM 󰑃  niri 26.04 (Wayland)
│ ├󰀻 󰑃  Flat-Remix-Blue-Dark [GTK3/4]
│ ├ 󰑃  Bibata-Modern-Ice (24px)
│ └ 󰑃  nvim
󰌢 SYSTEM 󰑃  MacBook Air (13-inch, Early 2015/2017) (1.0)
```
using this [imgcat](https://iterm2.com/utilities/imgcat)

cantona:
Fixes #1591 — `imgcat`, `icat`, `viu`, `chafa -f iterm` and sixel tools parsed fine but nothing ever rendered.

## Root cause

Since the kitty overlay rewrite, atlas graphics (sixel / iTerm2 OSC 1337) are parsed and uploaded into `sugarloaf.graphics` — a store **no draw path reads** (`Renderer::prepare` receives it as `_graphics`), and the per-cell `GraphicCell` references aren't consumed anywhere either. Only kitty images render, via `image_data` + `GraphicOverlay`s.

## Fix (3 commits)

1. **crosswords: make per-cell graphics state survive large inline images** — every covered cell used to allocate its own slot in the u16-indexed extras table, so ~a dozen screen-sized images exhausted the id space (images then rendered partially or not at all until restart). One shared slot per image row + a mark-and-sweep of unreferenced slots (which also reclaims images dropped off scrollback), plus two snapshot-freshness fixes.
2. **deps: enable TIFF decode** — chafa's iTerm2 renderer transmits TIFF; same change as #1428.
3. **rioterm: render sixel and iTerm2 graphics through the image overlay pipeline** — upload atlas pixels into the per-image texture store under a namespaced key (high bit keeps sequential `GraphicId`s out of the kitty protocol id space) and build overlays per frame by scanning visible rows for graphic cells, mirroring the kitty Unicode-placeholder pass. Row-runs coalesce into one overlay with a normalized source-rect slice, so partial overwrites hide exactly the erased region and images scroll with the grid. Also marks the panel dirty when `UpdateGraphics` arrives so the pixels aren't invisible until the next unrelated repaint.

## Tested (Linux/Wayland, Vulkan)

- `imgcat` (PNG), `chafa -f iterm` (TIFF), `chafa -f sixels`, `viu` — all render; `chafa -f kitty` unaffected.
- Repeated large (1600×900) images well past the old ~14-image exhaustion point; scrollback, `clear`, and partial overwrite behave.
- `cargo fmt`, `clippy`, and the full test suite pass; new behavior covered by updated graphics tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1493,6 +1493,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
+name = "fax"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "caf1079563223d5d59d83c85886a56e586cfd5c1a26292e971a0fa266531ac5a"
+
+[[package]]
 name = "fdeflate"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2144,6 +2150,7 @@ dependencies = [
  "moxcms",
  "num-traits",
  "png 0.18.1",
+ "tiff",
  "zune-core",
  "zune-jpeg",
 ]
@@ -4768,6 +4775,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f60246a4944f24f6e018aa17cdeffb7818b76356965d03b07d6a9886e8962185"
 dependencies = [
  "cfg-if 1.0.4",
+]
+
+[[package]]
+name = "tiff"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b63feaf3343d35b6ca4d50483f94843803b0f51634937cc2ec519fc32232bc52"
+dependencies = [
+ "fax",
+ "flate2",
+ "half",
+ "quick-error",
+ "weezl",
+ "zune-jpeg",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,8 @@ image_rs = { package = "image", version = "0.25.10", default-features = false, f
     "pnm",
     "webp",
     "bmp",
+    # chafa's iTerm2 renderer emits TIFF payloads (OSC 1337)
+    "tiff",
 ] }
 regex = "1.12.3"
 onig = { version = "6.5.1", default-features = false }
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -319,17 +319,22 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     }
                 }
             }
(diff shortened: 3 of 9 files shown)


cantona:
Hi @raphamorim 👋

First off — thanks for rio, it's become my daily driver. I've been scratching my own itches over the past weeks and have accumulated a batch of PRs. No rush at all, but when you find time, here's a suggested order to make reviewing painless:

**Bug fix for a known issue:**
- #1703 — sixel/iTerm2 inline images never render (fixes #1591). Root cause was graphics being uploaded to a store no draw path reads.

**Small, low-risk fixes (each a focused diff, quick to review):**
- #1678, #1679, #1680, #1682, #1684, #1687, #1688, #1689 — hot-reload, pixel-snapping, tab-bar and bindings fixes

**Small features:**
- #1681, #1683 (tab bar font-size/height config), #1690 (togglemaximized action)

**Bigger features (understand if these need more discussion — happy to adjust scope, split them up, or rework anything to fit your vision for rio):**
- #1694 — terminal triggers (iTerm2-style regex → action rules)
- #1704 — session save/restore (tabs, splits, CWDs, scrollback)

All are green on fmt/clippy/tests and rebased on current main. If any conflict by the time you get to them, ping me and I'll rebase same-day. And if some don't fit the project direction, no hard feelings — just close them.

Thanks for all the work you put into this project! 🙏

raphamorim:
Closing this one: sixel and iTerm2 rendering was rebuilt on main in #1729 and #1733, with images living as grid-anchored placements rather than overlays, and #1591 is closed. Two commits from this branch were merged into main with your authorship preserved (c6c3675937 making per-cell graphics state survive large inline images, and d1adb13728 enabling TIFF decode). Thanks a lot for the work and the diagnosis here, it moved this whole area forward. If you spot regressions in the new path please open an issue.

PR #1704, opened 2026-07-08
feat: session save/restore — tabs, splits, working dirs, scrollback

cantona:
Adds browser-style session save/restore: on exit rio can save the open tabs — including each tab's split layout, every pane's working directory, and the styled scrollback text — and offer to resume it on the next launch.

## Behavior

New `[session]` config table (default `never`, zero change unless enabled):

```toml
[session]
restore = "prompt"          # never | prompt | always
max-scrollback-lines = 2000
```

- **`prompt`** — quitting asks `save session? yes (y) / no (n)` (Esc aborts the quit); launching with a saved session asks `resume last session?`. The implicit slot is consumed once restored, so a stale session never re-prompts.
- **`always`** — saves on every exit (all windows) and restores on launch, silently.
- **Save on demand** — `ctrl+shift+s` (`cmd+shift+s` on macOS), bindable as `savesession`; flashes a transient "session saved" notice.
- **Named sessions** — `rio --session work` binds an instance to `sessions/work.json` (restored on launch, saved back on exit); the command palette gains `Save Session As...` (type a new name or overwrite an existing one) and `Restore Session...` (fuzzy picker; appends the session's tabs to the current window). Named sessions are workspaces and are never auto-deleted.

Restored panes reopen as fresh shells at their saved directory with the previous styled output repainted above the prompt — like a browser restoring URLs, not page state; running programs are not resurrected.

## Implementation

- **`Crosswords::scrollback_to_ansi`** serializes history + screen into text with minimal SGR runs (style diffed cell-to-cell, bg-only cells emit their inline color, zero-width combiners preserved); restore replays it through a `Processor` into the fresh terminal. Covered by a dump→replay→compare round-trip test.
- **Capture** walks each `ContextGrid`'s taffy tree (container flex direction + child `flex_grow` give the split shape and ratios) into a serde model; CWD uses `foreground_process_path` with OSC 7 fallback — the same priority new-tab inheritance uses.
- **Restore** replays splits via `split_right`/`split_down`, re-applies child weights, and spawns each pane through the spawn pty path with the saved directory (the fork path has no chdir, so restored panes force `use_fork = false`). Restored tabs follow the same margin choreography as `create_tab`, and when the platform applies the requested window size immediately (Wayland emits no `Resized` then) the resize pipeline runs inline.
- Prompts reuse the `ConfirmQuit` overlay pattern; `confirm-before-quit` chains into the save prompt.

## Limitations (documented)

- Windows: per-pane CWD capture unavailable (`foreground_process_path` unimplemented there).
- Window position restore is a no-op on Wayland (compositor-owned placement); size restores everywhere.
- Closing the last shell with `exit` ends rio without a save prompt — saves trigger on quit actions.

Tested on Linux/Wayland: tabs + right/down splits with adjusted ratios, colored scrollback (`ls --color`, git log), all three modes, named sessions via CLI and palette, repeated save/restore cycles. `cargo fmt`, `clippy`, and the full test suite pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4049,6 +4049,7 @@ dependencies = [
  "rio-window",
  "rustc-hash 2.1.2",
  "serde",
+ "serde_json",
  "smallvec",
  "sugarloaf",
  "taffy",
--- a/frontends/rioterm/Cargo.toml
+++ b/frontends/rioterm/Cargo.toml
@@ -42,6 +42,7 @@ image_rs = { workspace = true }
 libc = { workspace = true }
 parking_lot = { workspace = true }
 serde = { workspace = true }
+serde_json = "1.0"
 taffy = { version = "0.10.1", features = ["flexbox","grid"] }
 teletypewriter = { workspace = true }
 unicode-width = { workspace = true }
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -36,6 +36,7 @@ pub struct Application<'a> {
     router: Router<'a>,
     scheduler: Scheduler,
     app_id: Option<String>,
+    session_name: Option<String>,
 }
 
 impl Application<'_> {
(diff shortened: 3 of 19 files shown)


olkva:
I can confirm that feature is working, but it needs a rebase with corresponding resolving of merge conflicts. I have build rebased version of rio using this fork, since this is exactly what i was missing.

cantona:
> I can confirm that feature is working, but it needs a rebase with corresponding resolving of merge conflicts. I have build rebased version of rio using this fork, since this is exactly what i was missing.
> 
> 

I actually have my own fork version in my repo, you can try if interest.

Issue #1314, opened 2025-09-28
Rio renders images at incorrect sizes


og900aero:
I tried Rio, and I noticed that when I use the Yazi file manager and scroll through images, the preview image sometimes slips and jumps a line. 

https://github.com/user-attachments/assets/12380e1c-57cd-4b0b-a75e-c65a521e737f

I spoke with the developer of Yazi (also written in Rust), and as he examined it, he found that Rio renders images at incorrect sizes.

TheVimDev:
This happens to me as well. I'm daily driving Rio, but when this happens it tends to disrupt my workflow.
Thank you for such an amazing terminal, tho

ecocode:
confirm this issue

raphamorim:
The kitty graphics placement rework in #1726 (part of the upcoming 0.4.12) fixed several bugs in this area that are likely behind what Yazi hits: native-size images were stretched vertically to their cell box (up to a full cell) after any font or window resize, cell metrics went stale after DPI-only changes, and `x=`/`y=` source crops moved the placement instead of cropping. Yazi uses the kitty protocol on Rio, so a retest once 0.4.12 is out would be very helpful. If it still reproduces, a note about which preview protocol Yazi picked (`yazi --debug` shows it) would narrow it down.

raphamorim:
The image placement math was rebuilt in #1726 and #1729 (both on main, shipping in 0.4.12). Could you retest on main and report whether this is still reproducible?

raphamorim:
Closing for now but can reopen anything

og900aero:
I try the following:
1. git clone --depth=1 https://github.com/raphamorim/rio.git
2. cargo build -p rioterm --release --no-default-features --features=wayland
3. run with: target/release/rio
4. run yazi, and I see this in preview area:

<img width="1908" height="218" alt="Image" src="https://github.com/user-attachments/assets/585a308a-92c4-4f7a-b81d-77be9bda63e8" />

I am on Debian 13 with waylan (MangoWM)

raphamorim:
thanks, @og900aero fixed in #1736

raphamorim:
[...]

Issue #1293, opened 2025-09-02
Images are drawn ~0.5 lines higher than they should


itsjunetime:
Here's a screenshot:

<img width="760" height="309" alt="Image" src="https://github.com/user-attachments/assets/7e392ff9-6bed-442f-b91a-c5a4b89aa751" />

This seems to be consistent with every app that I use to display images, and, unless I'm mistaken, the image should be displayed like 0.5 lines lower than it's actually being displayed at.

The screenshot was taken from a build on current main.

I tried to figure this one out myself but I was just running in circles and couldn't figure out exactly where this draw location is actually determined.

itsjunetime:
Even after my fixes in #1340, it panics on startup. Here's the message:

```
thread 'main' panicked at /home/june/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:1365:26:
wgpu error: Validation Error

Caused by:
  In Device::create_render_pipeline
    Error matching ShaderStages(VERTEX) shader requirements against the pipeline
      Location[4] Float32 interpolated as Some(Perspective) with sampling Some(Center) is not provided by the previous stage outputs
        Input is not provided by the earlier stage in the pipeline


stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:75:14
   2: wgpu::backend::wgpu_core::default_error_handler
             at /home/june/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:693:5
   3: wgpu::backend::wgpu_core::ErrorSinkRaw::handle_error_or_return_handler
             at /home/june/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:677:21
   4: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error_inner
             at /home/june/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-27.0.1/src/backend/wgpu_core.rs:305:18
   5: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error
[...]

PR #1726, opened 2026-07-21
fix kitty source crop handling and test overlay geometry

raphamorim:
First PR from the image placement audit (#1293/#1314/#1276/#1441 investigation). Six placement fixes, per frame crop resolution, CPU compositing, and the test harness they all run under.

## Bug 1: kitty `x=`/`y=` moved the placement instead of cropping the image

Per the kitty graphics spec a placement always renders at the cursor; `x=`/`y=`/`w=`/`h=` select the source rectangle within the image. Rio used `x=`/`y=` as the destination column/row, and the stored source rectangle was never consumed by the render path. Any client that tiles or crops images got both a teleported placement and an uncropped image.

## Bug 2: wgpu sampled wrong texture coordinates for sliced placements

The image shaders disagreed on `source_rect` semantics: Metal and Vulkan treat it as `[u0, v0, u1, v1]` (origin, end) but the wgsl shader computed `xy + zw * corner` (origin, size), so on wgpu any placeholder run past an image's first row sampled the wrong region. All three backends now share the origin/end convention.

## Bug 3: native-size images stretched to their cell box on resize

The resize path recomputed every placement's pixel size as `columns * cell_width`, guarded by `columns > 0` — always true for derived spans. Any font or window change inflated native images by up to a full cell (the "incorrect sizes" signature from #1314/#1441). Placements now record the requested `c=`/`r=` separately; native ones keep their pixel size.

## Bug 4: font/DPI change with unchanged cols/rows never rescaled anything

`Crosswords::resize` early-returns before `graphics.resize` when grid dims are unchanged, leaving cell metrics stale after DPI-only changes. The early-return path now syncs cell size and rescales placements.

## Bug 5: sub-cell offsets were destroyed by shrinking the cell size

Offsets were clamped destructively at place/rescale time. They're now stored raw and clamped where read, so repeated resizes are lossless.

## Bug 6: unstable paint order for equal z-indexes

Placements were sorted by z alone with ties in hash map iteration order (frame to frame flicker potential). The frontend tie-breaks on `(image_id, placement_id)`; the CPU flatten relies on that order via a stable sort.

## Per frame crop and size resolution

The crop and display size are no longer baked into the placement at place time. `KittyPlacement` stores the raw request (`x=`/`y=`/`w=`/`h=`, zero meaning "to the edge", plus the requested cell span) and the render path resolves everything against the image's current dimensions and cell size every frame. Consequences, each covered by a test:

- A retransmission (`a=t`) that changes an image's dimensions re-clamps existing placements instead of showing a stale crop, refreshes their grid footprint, and pushes the new pixels so the display updates without a re-place.
- A crop fully outside the image stores the placement (kitty answers OK), renders nothing, occupies no cells, and leaves the cursor untouched — and becomes visible if a later retransmit makes the image large enough.
- Virtual (Unicode placeholder) placements honor the source crop too: the crop, not the whole image, is aspect-fitted into the placement grid, and per run slices map into full image texture coordinates. No crop degenerates to the exact previous math.

## Cursor movement matches kitty

`C=0` used to leave the cursor at column 0 of the image's last row, so the next prompt overwrote the image's bottom row. It now linefeeds once per occupied row (scrolling the image fully into view) and lands on the image's last row at the first column after it, clamped to the grid edge — kitty's behavior. Invisible placements don't move the cursor.

## Feature: CPU backend composites kitty images

The CPU (softbuffer) backend skipped image overlays entirely. It now composites them with the same three layer z ordering as the GPU backends via a nearest neighbor, alpha blended blit honoring the normalized source rectangle, with overlays folded into the frame skip hash. This is also what makes image rendering testable without a GPU.

## Tests

- Placement geometry (rio-backend, 20+ new): crop at cursor, raw storage with read time clamping, span derivation, one axis and both axes scaling, scrollback positioning, culling, partial visibility, native vs cell-sized rescale, retransmit re-clamping and degenerate-to-visible transitions, cursor movement at the screen bottom with scrollback, virtual run crops with letterbox culling.
- Pixel goldens (sugarloaf, pure CPU): exact pixel 1:1 blit, crop selection, nearest neighbor upscale, alpha blend, edge clipping, transparent passthrough.

579 rio-backend, 152 sugarloaf, 162 rioterm tests pass.

### DIFF
--- a/frontends/rioterm/src/renderer/mod.rs
+++ b/frontends/rioterm/src/renderer/mod.rs
@@ -414,31 +414,40 @@ impl Renderer {
                 overlays.clear();
 
                 if has_overlays {
-                    let history_size = rc.history_size as i64;
-                    let display_offset = rc.display_offset as i64;
-                    let screen_lines = rc.screen_lines as i64;
+                    let viewport = rio_backend::ansi::graphics::OverlayViewport {
+                        cell_width,
+                        cell_height,
+                        origin_x,
+                        origin_y,
+                        history_size: rc.history_size as i64,
+                        display_offset: rc.display_offset as i64,
+                        screen_lines: rc.screen_lines as i64,
+                    };
 
                     for p in &rc.kitty_placements {
-                        let screen_row = p.dest_row - (history_size - display_offset);
-                        let image_bottom_row = screen_row + p.rows as i64;
-                        // Cull only if fully off-screen (like )
-                        if image_bottom_row <= 0 || screen_row >= screen_lines {
+                        let (image_width, image_height) = rc
+                            .kitty_images
+                            .get(&p.image_id)
+                            .map(|stored| (stored.data.width, stored.data.height))
+                            .unwrap_or((0, 0));
+                        let Some(geometry) =
+                            rio_backend::ansi::graphics::kitty_overlay_geometry(
+                                p,
+                                image_width,
+                                image_height,
+                                &viewport,
+                            )
+                        else {
                             continue;
-                        }
+                        };
                         overlays.push(rio_backend::sugarloaf::GraphicOverlay {
                             image_id: p.image_id,
-                            // kitty X=/Y= offset, supports sub-cell positioning
-                            x: origin_x
-                                + p.dest_col as f32 * cell_width
-                                + p.cell_x_offset as f32,
-                            y: origin_y
-                                + screen_row as f32 * cell_height
-                                + p.cell_y_offset as f32,
-                            width: p.pixel_width as f32,
-                            height: p.pixel_height as f32,
+                            x: geometry.x,
+                            y: geometry.y,
+                            width: geometry.width,
+                            height: geometry.height,
                             z_index: p.z_index,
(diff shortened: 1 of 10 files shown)


raphamorim:
Latest commit addresses two issues found in a further review pass against the reference implementation:

- `C=0` cursor movement now discards a pending wrap. The old path cleared it implicitly through `carriage_return`; the rewritten movement assigned the position directly, so a character printed into the last column right before a placement made the following caption wrap one row below where it belonged.
- `a=T` (transmit and display) now refreshes the grid footprint of the image's other live placements, symmetric with the `a=t` path. Without it, a placement's cached cell span could disagree with what renders after a retransmit that changed the image's dimensions (early scroll cull, delete-by-intersection misses).

Also unified cell stride rounding across the footprint and render paths (they could only diverge with fractional strides, which don't occur today), and documented that `pixel_width`/`pixel_height` are footprint bookkeeping the render path never reads. Regression tests added for both fixes. 581 rio-backend tests pass.

PR #1728, opened 2026-07-21
restore sixel and iterm2 rendering with texture lifecycle fixes

### LINKED ISSUE #1591: iTerm2 image protocol broken on 0.4.2 & 0.4.3

raphamorim:
Builds on #1703 by @cantona — the three commits from that PR are preserved here with original authorship, rebased onto main (over the kitty placement rework from #1726), plus fixes from a performance and correctness review on top.

## From #1703 (rebased, unchanged)

- Shared extras slot per covered image row so large sixel/iTerm2 images stop exhausting the u16 extras id space, with a mark-and-sweep reclaim when the table nears the cap.
- TIFF decode for iTerm2 payloads (chafa emits TIFF).
- Atlas graphics (sixel/iTerm2) routed through the per-image texture overlay pipeline: visible rows are scanned for graphic cells and contiguous runs coalesce into one overlay per image row with a normalized source slice, so partial overwrites hide exactly the erased region and images scroll with the grid.

Fixes #1591.

## Added on top

**Disjoint image key namespaces.** The bridge keyed atlas textures as `0x8000_0000 | id` in the shared u32 image store — but kitty clients may pick any u32 image id (kitten icat uses random ones), so half the kitty id space collided with atlas textures, and the two would silently clobber each other. Image keys are now u64: kitty ids map verbatim (the full u32 space belongs to the client), atlas graphics live above 2^32.

**GPU texture eviction.** The per-image texture cache was only ever cleared wholesale; removing an image never freed its texture. Atlas `GraphicId`s are sequential and never reused, so sixel animation (timg, mcat) uploaded a new GPU texture per frame, forever. Removals now flow as final image keys through one queue (`sugarloaf.remove_image` frees both the pixel store and the cached texture), and kitty evictions — which previously pushed their ids into the removal queue where they were misinterpreted as atlas keys — now evict the right entry.

**Tests.** Atlas overlay scan unit tests (run coalescing, erased-cell splitting with positionally derived slices, font-resize scaling), key namespace disjointness, and the TextureRef drop queue.

## Performance notes from the review

- The per-frame row scan only runs on frames that render at all (Rio renders on damage), skips rows without extras via the existing per-row flag, and coalesces runs — measured cost is well bounded.
- The Noop/CursorOnly damage fall-through (snapshot freshness fix from #1703) is a per-row boolean check with early exit — negligible, kept as is.
- The texture eviction above was the real lifecycle cost: unbounded VRAM growth under any repeated-image workload.

583 rio-backend, 152 sugarloaf, 165 rioterm tests pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1502,6 +1502,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "da7c62ceae207dd37ea5b845da6a0696c799f85e97da1ab5b7910be3c1c80223"
 
 [[package]]
+name = "fax"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "caf1079563223d5d59d83c85886a56e586cfd5c1a26292e971a0fa266531ac5a"
+
+[[package]]
 name = "fdeflate"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2170,6 +2176,7 @@ dependencies = [
  "moxcms",
  "num-traits",
  "png 0.18.1",
+ "tiff",
  "zune-core",
  "zune-jpeg",
 ]
@@ -4829,6 +4836,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1ad99c4c6d32803332c548b1af0540b357b3f5fc0be8f6c6bfe8b2e6ae784070"
 dependencies = [
  "cfg-if 1.0.4",
+]
+
+[[package]]
+name = "tiff"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b63feaf3343d35b6ca4d50483f94843803b0f51634937cc2ec519fc32232bc52"
+dependencies = [
+ "fax",
+ "flate2",
+ "half",
+ "quick-error",
+ "weezl",
+ "zune-jpeg",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


raphamorim:
Superseded by #1729, which renders sixel/iTerm2 as grid content (resident per-row image runs built during the damage-driven row rebuild) instead of bridging them into the overlay pipeline with a per-frame cell scan. The texture lifecycle fixes from this PR (key namespaces, GPU texture eviction) carried over.

PR #1747, opened 2026-07-22
expand and validate working-dir, honor it on fork path

### LINKED ISSUE #1088: Report invalid working-dir paths
This config file tends to crash due to wrong invalid path, since doesn't evaluate relative paths

```
theme = "catpuccin"
working-dir = "~"
confirm-before-quit = false

[editor]
program = "nvim"

[fonts]
family = "Aporetic Sans Mono"
size = 14
```

raphamorim:
Three related `working-dir` problems (#1088):

- `working-dir = "~"` (or `~/projects`) was passed to the pty literally and failed, since nothing expanded the tilde.
- A nonexistent directory made the spawn fail or land somewhere unexpected with no feedback.
- The fork pty path, the default on Linux/BSD via `use-fork = true`, never received the working directory at all, so the option was silently ignored there (same family as the `shell.args` bug fixed in #1721).

`resolve_working_dir` now expands a leading tilde, validates the path is a directory, and logs a warning while falling back to inheriting the current directory when it is not. The spawn path resolves through it, and `create_pty_with_fork` takes the directory and chdirs in the forked child before exec (resolved in the parent so the failure path can log). Six unit tests cover the resolution rules.

Closes #1088.

### DIFF
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -263,6 +263,7 @@ impl<T: EventListener + Clone + std::marker::Send + 'static> ContextManager<T> {
                 pty = match create_pty_with_fork(
                     &Cow::Borrowed(&config.shell.program),
                     &config.shell.args,
+                    &config.working_dir,
                     cols,
                     rows,
                     initial_winsize.width,
--- a/teletypewriter/examples/spawn.rs
+++ b/teletypewriter/examples/spawn.rs
@@ -10,7 +10,7 @@ fn main() -> std::io::Result<()> {
     use teletypewriter::{create_pty_with_fork, ProcessReadWrite, Pty};
 
     let shell = Cow::Borrowed("bash");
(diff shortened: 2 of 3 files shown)


raphamorim:
Review notes on this one: the fork-side chdir is async-signal-safe with the CString built pre-fork, defaults are unaffected (default working-dir is None), and all three unix spawn entry points are covered. One regression the first revision would have introduced: the flatpak branch forwards `--directory` to `flatpak-spawn --host`, where the host may see directories the sandbox cannot, so validating inside the sandbox could silently drop a working host path. Reworked so tilde expansion applies everywhere but existence validation happens only at the local use sites (the builder cwd and the fork child); the flatpak host keeps receiving the expanded path untouched. The failing Build check is the repo-wide nix/FlakeHub credential issue, same as every sibling PR.

mirsella:
hello ! LGTM, anything blocking ? can adopt it if anything is missing

PR #1750, opened 2026-07-23
feat: make the tab strip fully configurable (height, font size, geometry, width, fills)

cantona:
Consolidates the tab-strip configurability work into a single feature and
supersedes the three narrower PRs #1681, #1683 and #1682.

The floating-island tab strip hard-codes its gap, vertical inset, corner
radius, height, title font size and per-tab width cap, and derives the
island fills from the window background's luminance — the configured tab
colors only tint the titles. This exposes all of it under `[navigation]`:

- `tab-font-size` — tab title font size
- `tab-bar-height` — strip height
- `tab-max-width` — per-tab width cap (`0` removes the cap so tabs expand
  to share the whole strip); replaces the previous `max-tab-width`
- `tab-gap`, `tab-inset-y`, `tab-radius` — strip geometry (zeroing all
  three gives a classic flat, full-height strip; defaults keep the stock
  look)
- `tab-fill` / `tab-fill-active` — override the adaptive luminance fills
  with fixed colors; unset keeps the current behavior

Geometry moves into a `TabGeom` carried by the island and threaded through
`island_rect`, the strip layout and the close-button hit tests, so custom
values stay consistent between painting and mouse handling. All values
apply on config hot-reload.

Supersedes #1681, #1683, #1682.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1168,9 +1168,11 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                         && button == MouseButton::Left
                         && route.window.screen.allow_manual_dragging
                     {
-                        use crate::renderer::island::ISLAND_HEIGHT;
                         let scale = route.window.screen.sugarloaf.scale_factor();
-                        if route.window.screen.mouse.y <= (ISLAND_HEIGHT * scale) as f64 {
+                        let tab_bar_height =
+                            route.window.screen.renderer.navigation.tab_bar_height;
+                        if route.window.screen.mouse.y <= (tab_bar_height * scale) as f64
+                        {
                             let _ = route.window.winit_window.drag_window();
                         }
                     }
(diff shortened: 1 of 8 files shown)


raphamorim:
that's a cool idea, could you share some screenshots with variants? 
thank you!

cantona:
Sure! Here are some variants — all the same build, switched live by editing `[navigation]` and saving (every value hot-reloads).

tab strip variants

Top to bottom: stock defaults; `tab-gap`/`tab-inset-y`/`tab-radius = 0` for a classic flat, full-height strip; `tab-max-width = 0` plus fixed `tab-fill`/`tab-fill-active` for a browser-style strip that shares the full width; and a roomier look with `tab-bar-height = 46`, `tab-font-size = 14` and larger gap/radius/inset.

A few more angles — density extremes, fixed accent fills, full-width sharing under 8 tabs, and (with no `[navigation]` config at all) the adaptive fills tracking a light theme:

more variants

Full window on the light theme, all defaults — the strip still derives its fills from the background's luminance, so existing setups look exactly the same until someone opts in:

light theme

olkva:
Very nice feature! It would also be great to add possibility to show tabs below (under the terminal). Example from tabby terminal: 
image

raphamorim:
I think this feature is great too, just wondering if we can make more customisation available

PR #1751, opened 2026-07-23
fix: recompute grid rows when the tab bar appears/disappears

cantona:
Creating a tab toggles hide_if_single, which grows scaled_margin.top to
make room for the tab bar. update_dimensions then re-laid out via
apply_taffy_layout, but never refreshed Taffy's available height for the
new margin -- so the grid kept its old (taller) row count and the bottom
row was clipped off the window until a window resize (which calls
try_update_size) corrected it. Refresh the Taffy size with the current
scaled_margin in update_dimensions before laying out.

### DIFF
--- a/frontends/rioterm/src/layout/mod.rs
+++ b/frontends/rioterm/src/layout/mod.rs
@@ -1206,6 +1206,15 @@ impl<T: rio_backend::event::EventListener> ContextGrid<T> {
             dim.update_dimensions(text_dims, cell);
         }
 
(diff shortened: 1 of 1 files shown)


cantona:
Redundant now — this is fixed on main by #1723 (`update_scaled_margin` re-runs `try_update_size` whenever the margins change, which covers the tab-bar band growing/shrinking on the single-tab toggle). Same root cause, resolved at the margin-mutation point. Closing.

PR #1754, opened 2026-07-23
fix(performer): disarm sync-update timeout on inline ESU

### LINKED ISSUE #1753: Typing echo is delayed ~150 ms in synchronized-output TUIs through ConPTY
## Environment

- Rio for Windows using ConPTY
- WSL shell running inside Rio
- Codex CLI 0.145.0 as a real-world reproducer

The issue was not reproduced on macOS, where ConPTY is not involved.

## Summary

Typing in TUIs using synchronized output can be displayed roughly 150–200 ms
late when Rio runs through ConPTY.

Keyboard input reaches the application immediately and the application emits
its updated frame quickly, but Rio buffers the frame until its 150 ms
synchronized-update timeout expires.

Plain applications such as `cat` are not affected.

## Reproduction

Run these commands in a WSL shell inside Rio for Windows (`Ctrl-C` to stop):

```bash
# Control: immediate echo without synchronized output.
while IFS= read -rsn1 key; do
  printf '\rWithout 2026: %s' "$key"
done

# Reproduction: one write containing BSU, content and ESU.
while IFS= read -rsn1 key; do
  printf '\e[?2026h\rWith 2026: %s\e[?2026l' "$key"
done
```

Expected: both loops update immediately.

Actual: the second loop displays each isolated key roughly 150–200 ms late.

Codex is a concrete real-world reproducer. Its TUI depends on Ratatui and
Crossterm, and its draw path explicitly wraps frames in
[`sync_update`](https://github.com/openai/codex/blob/rust-v0.145.0/codex-rs/tui/src/tui.rs#L895-L951).

## Investigation

Instrumentation showed:

- input reached the application in about 1–2 ms;
- the application emitted its updated frame after about 20 ms;
- the visible update took about 206 ms;
- `cat` remained immediate in the same Rio window.

ConPTY was observed returning the synchronized-update markers together, ahead
of the frame content:

```text
chunk 1: ESC[?2026h ESC[?2026l
chunk 2: frame content
```

Rio arms a 150 ms timeout when parsing `ESC[?2026h`. When the matching
`ESC[?2026l` is parsed inline through the normal CSI dispatch path, the timeout
is not cleared. The following frame content is therefore treated as belonging
to a synchronized update and remains buffered until the timeout expires.

## Proposed fix

Clear the synchronized-update timeout when dispatching `?2026l`, mirroring the
existing `?2026h` behavior.

When replaying a synchronized buffer that contains a new BSU after the current
ESU, explicitly re-arm the timeout for that remaining update.

Regression tests cover:

- an inline BSU+ESU pair followed by frame content;
- an ESU followed by a new BSU;
- the existing split-chunk ESU path.

AI assistance: the diagnosis and fix were developed with Claude Code using
Fable 5.

cmoron:
Closes #1753

## Summary

- clear the synchronized-update timeout when a mode 2026 ESU is parsed inline;
- re-arm it when replaying the buffer leaves a new BSU pending;
- add regression coverage for inline, nested and split ESU paths.

## Root cause

ConPTY was observed returning a synchronized update's BSU+ESU pair together,
ahead of the frame content that originally sat between them.

Rio arms a 150 ms timeout when parsing the BSU. Since the inline ESU goes
through the normal CSI dispatch path rather than `advance_sync`, it did not
clear that timeout. The following frame content was consequently buffered until
the timeout expired.

The ESU dispatch now mirrors the BSU dispatch by clearing the timeout. If a new
BSU remains after replaying the synchronized buffer, `stop_sync_internal`
explicitly re-arms it.

## Testing

- `cargo test -p rio-backend --release sync_ -- --nocapture`
- `cargo test -p rio-backend --release` — 620 passed
- `cargo clippy -p rio-backend --lib --all-features -- -D warnings`
- `cargo build -p rioterm` on macOS arm64
- manual Windows/ConPTY validation
- manual macOS smoke test; the original issue is not reproduced without ConPTY

Local `make test` remains blocked by Rust 1.96 warnings in unchanged upstream
tests under `sugarloaf` and `crosswords`.

AI assistance: the diagnosis and fix were developed with Claude Code using
Fable 5.

### DIFF
--- a/rio-backend/src/performer/handler.rs
+++ b/rio-backend/src/performer/handler.rs
@@ -598,6 +598,15 @@ impl Processor {
                 let new_len = self.state.sync_state.buffer.len() - bsu_offset;
                 self.state.sync_state.buffer.copy_within(bsu_offset.., 0);
                 self.state.sync_state.buffer.truncate(new_len);
+
+                // The bytes we just replayed above may contain an inline
+                // ESU, which (see the `('l', [b'?'])` dispatch arm) clears
+                // the timeout. But a new BSU is still open in what's left
+                // of the buffer, so the sync must stay pending.
+                self.state
+                    .sync_state
+                    .timeout
+                    .set_timeout(SYNC_UPDATE_TIMEOUT);
             }
             // Report mode and clear state if no new BSU is present.
             None => {
@@ -1318,6 +1327,17 @@ impl<U: Handler> Perform for Performer<'_, U> {
             }
             ('l', [b'?']) => {
                 for param in params_iter.map(|param| param[0]) {
+                    // ConPTY can re-emit a synchronized update's BSU+ESU
+                    // pair glued together in one chunk, ahead of the frame
+                    // content that was actually between them. That means
+                    // the ESU can be parsed here, inline, instead of going
+                    // through `advance_sync`/`stop_sync_internal`. Without
+                    // clearing the timeout, it stays armed and all
+                    // subsequent content gets buffered until it expires.
+                    if param == NamedPrivateMode::SyncUpdate as u16 {
+                        self.state.sync_state.timeout.clear_timeout();
+                    }
+
                     handler.unset_private_mode(PrivateMode::new(param))
                 }
             }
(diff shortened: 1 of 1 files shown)


cmoron:
Superseded by #1789, which landed the inline ESU timeout fix plus the follow-up flush behavior in the new `rio-vt` architecture. Closing this older implementation to avoid keeping a duplicate PR open.

PR #1758, opened 2026-07-23
fix(config): keep the live config when a hot-reload fails to parse

cantona:
fix(config): keep the live config when a hot-reload fails to parse

## Problem

The `UpdateConfig` (hot-reload) handler did:

```rust
let (config, config_error) = match Config::try_load() {
    Ok(config) => (config, None),
    Err(error) => (Config::default(), Some(error)),   // <- substitutes defaults
};
...
self.config = config;   // applied, defaults and all
```

So when a `config.toml` typo is saved mid-edit, the live config is
**replaced with `Config::default()`** — fonts, colors, and key bindings
all snap back to defaults until the file parses cleanly again. The error
screen does show, but it renders over a terminal that has lost the
user's whole configuration.

## Fix

On a failed reload, keep running on the current config: report the parse
error to every window, request a redraw, and skip the reconfiguration
entirely. The next successful save applies normally and clears the error
screen.

Startup is intentionally unchanged — at initial load there is nothing
live to preserve, so a broken `config.toml` still falls back to defaults
there.

## Testing

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p rioterm --all-features` — no new warnings
- `cargo test --features wgpu` — all suites pass (0 failed)

Single-file change to the `UpdateConfig` handler; no config schema or
API changes.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -520,10 +520,24 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                 }
             }
             RioEventType::Rio(RioEvent::UpdateConfig) => {
-                let (config, config_error) = match rio_backend::config::Config::try_load()
-                {
-                    Ok(config) => (config, None),
-                    Err(error) => (rio_backend::config::Config::default(), Some(error)),
+                // A config.toml typo saved mid-edit must not reset the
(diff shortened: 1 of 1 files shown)


cantona (commented):
@raphamorim please consider this, if not, let me know what can be improved.

PR #1775, opened 2026-07-28
fix(x11): redraw windows after mapping

### LINKED ISSUE #1604: When switching workspace in i3 wm, rio does not redraw the window until manual key input
## Platform Details
OS: Linux Mint
DE/WM: i3
Display server: X11
Shell: bash
GPU: intel integrated
Compositor: picom
Rio version: 0.4.4

## Description
Lets say I have a workspace with Rio open - I switch to a different workspace (say a browser). When switching back to Rio's workspace, the window keeps the ghost pixels from the previous workspace. Only when I press a key the terminal window is shown again along with its contents.

I have tried it with picom compositor on and off - both same result so picom is probably not the issue. On the same set up, alacritty works fine out of the box without any issues.

Most likely when switching the workspace i3 emits an event or a hook which Rio is not catching.

## Expected Behaviour
Rio should repaint its terminal contents immediately when the i3 workspace becomes visible again.

## Actual Behaviour
Rio sometimes shows stale pixels from the previously visible workspace until keyboard input forces a redraw.

## Temporary Workaround
Setting `renderer.strategy` to "Game" has fixed it. But consider a cleaner fix so that it is compatible with the "events" strategy as well.
I have toggled both these options:
`renderer.disable-unfocused-render` and `renderer.disable-occluded-render` but it **did not help**.

0xNOY:
## Summary

- request a redraw whenever an X11 window is mapped
- restore event-driven rendering immediately after i3/XMonad workspace switches
- keep the existing redraw coalescing behavior when an `Expose` event follows

## Root cause

Tiling window managers such as i3 hide workspace windows with `UnmapWindow` and map
them again when the workspace becomes visible. A compositor can retain the old
window pixels while doing this, so X11 does not necessarily send an `Expose` event
after the map.

Rio already re-emits the current focus state from `MapNotify`, which marks the
terminal contents dirty. However, when that focus state has not changed, no redraw
is requested and the event-based renderer does not present another frame until
keyboard or mouse input produces one.

Queueing `RedrawRequested` immediately after the mapped window's focus event makes
the dirty frame present without switching to the continuously-rendering `Game`
strategy. If X11 also sends `Expose`, the existing redraw queue deduplicates both
requests for the same window.

Closes #1604.

## Validation

- `cargo fmt -- --check --color always`
- `cargo test -p rio-window` (17 tests and 18 doc tests passed; 4 doc tests ignored)
- `cargo check -p rioterm --no-default-features --features x11`

### DIFF
--- a/rio-window/src/platform_impl/linux/x11/event_processor.rs
+++ b/rio-window/src/platform_impl/linux/x11/event_processor.rs
@@ -856,6 +856,15 @@ impl EventProcessor {
         };
 
         callback(&self.target, event);
+        // Compositors can restore a mapped window without sending Expose, so
+        // event-driven renderers must explicitly present its retained contents.
+        callback(
+            &self.target,
+            Event::WindowEvent {
+                window_id,
+                event: WindowEvent::RedrawRequested,
+            },
+        );
 
         // Window is mapped → flip the mapped bit and (re)evaluate
         // the per-window vsync timer. We DON'T touch the obscured


0xNOY:
Superseded by #1776 after renaming the head branch to `fix/redraw-on-x11-map`.

PR #1788, opened 2026-08-01
fix(rio-vt): disarm sync-update timeout on inline ESU

marc2332:
its like https://github.com/raphamorim/rio/pull/1754/changes but done in the rio-vt crate so that it works with other emulators like freya-terminal https://github.com/marc2332/freya/pull/2116


marc2332:
I noticed this because when I exited neovim it would fall into the sync timeout I added in freya-terminal. Now its instant.

raphamorim:
moved this forward in #1789 — kept your commit as-is and added an embedder-facing follow-up on top (flush expired updates on the next advance, since freya/ratty/etc don't poll the timeout like rio's event loop does). will close this one once that lands, thanks again!

Issue #1791, opened 2026-08-02
[Windows] Closing a fullscreen window briefly flashes the windowed-size window


baoyu0:
## Summary

On Windows, when Rio is in fullscreen (toggled via F11 / `ToggleFullscreen`) and the window is closed (Alt+F4 or close button), a brief flash of the small windowed-size window (default 600×400) appears before the app exits.

## Steps to reproduce

1. Windows 11 Pro, build 10.0.26200
2. Rio 0.5.5 (latest stable), default install
3. Open Rio → press F11 to enter fullscreen
4. Close the window while still fullscreen (Alt+F4 or click the X)

Observed: for a frame or two, the pre-fullscreen windowed-size window is visible (restored size + decorations), then everything disappears.

Expected: the window closes directly from fullscreen with no windowed-size flash.

## Diagnosis / suspected cause

This looks like a winit-on-Windows behavior: on close while in fullscreen, the saved windowed rect/state is restored during teardown, and one frame of the windowed window gets painted before the window is destroyed. Rio does not hide the window before closing. Similar flash artifacts were fixed for startup in #640, this is the close-path equivalent.

## Environment

- OS: Windows 11 Pro 22H2+ (build 26200)
- Rio: v0.5.5 (rioterm 0.5.5)
- Config: F11 = `ToggleFullscreen`, `decorations = "Enabled"`, `margin = [8,8,8,8]`, window mode default (Windowed)
- GPU/renderer: wgpu/DX12 default

Happy to provide a screen recording if needed.

Issue #640, opened 2024-09-03
Flash of white during startup on Windows (may possibly affect other platforms)


fgimian:
Hey there, really great job on Rio.

On Windows, there's a rather unpleasant initial flash of white and black while the app is starting up. It may be worth keeping the window hidden until the first draw is completed.

Please see a recording below:

![Rio Startup](https://github.com/user-attachments/assets/06897c14-e7dd-40ec-ab9a-a06ce9ea50c5)

Cheers
Fotis

raphamorim:
Thank you for the issue!

Yes, I could reproduce it today. Will mark to take a look!

raphamorim:
Fixed on main, it was similar issue to https://github.com/vizia/vizia/pull/386

PR #1814, opened 2026-08-04
fix: remove Windows startup window flash

lalvarezt:
On Windows, Rio could show a blank native window before the first frame was ready, causing a brief flash and making startup feel slower. The fix keeps the window cloaked until the first frame is rendered, eliminating the blink and improving perceived startup speed.

Related to #640, #1791


raphamorim (approved):
thank you!

PR #1835, opened 2026-08-09
Render a hollow cursor in unfocused windows

mirsella:
An active panel kept drawing a solid cursor after its window lost focus because the cursor style only checked panel focus. It now checks the window too, so an unfocused window gets the hollow cursor as expected.

I tested this on my local build and it works.


raphamorim:
Thanks!

PR #1846, opened 2026-08-09
fix(ime): render preedit inline and stop Windows key-repeat from stalling redraws

raphamorim:
Supersedes #1557 (commits by @Tryanks and @shiena, rebased onto main's grid pipeline, plus review fixes). **Draft — deep review 2026-08-10 found blocking defects in both halves; punch list below.** CI is green, but CI cannot see any of these.

## What this PR does

- **Inline IME preedit rendering** for rioterm (wezterm-style block at the cursor with a caret beam) — rioterm currently renders no preedit at all.
- **Windows: redraws via a posted message** so IME key-repeat can't starve `WM_PAINT`, plus IME message-handshake perf changes.

## Blockers found by deep review (must fix before un-drafting)

**Rendering half:**
1. **The caret beam is invisible in its primary case** — it's drawn in `named_colors.cursor` on a block filled with the same color (grid_emit.rs:2340 vs 1108-1110). Arrow-keying between Japanese conversion segments — the reason the caret exists — shows nothing.
2. **Bottom-row truncation** — the overlay wraps *down* and breaks at the last row (renderer/preedit.rs:82-98). A shell prompt on the bottom row (the dominant case) composing CJK near the right edge loses the tail and the caret. Needs the alacritty/wezterm slide-left fallback.
3. **Scrollback mid-composition** — `cursor.state.pos` is unshifted by display offset, and preedit forces a Block cursor *before* the visibility check (grid_emit.rs:527-531), so scrolling back paints the overlay + cursor over history at a stale position. Should suppress when `display_offset > 0`.
4. **Stale preedit across split/tab switches** — `Ime::Preedit` mutates only the current context and nothing clears it on switch; the old split keeps a frozen block *and* swallows Enter/Backspace/Ctrl-C on refocus (screen/mod.rs:694-696) until a new composition clears it. Needs preedit-clear on context switch.
5. **`Ime::Disabled` clears preedit without damage/redraw** (application.rs:1983-1991) — switching input source mid-composition leaves a ghost overlay.
6. **Grapheme clusters unrepresentable** — one-char-per-cell layout gives ZWJ-emoji three emoji plus tofu cells, and decomposed `e`+U+0302 renders as two cells. Needs grapheme segmentation in the overlay model.

**Windows half:**
7. **Lost redraw in the buffered branch** — the `REDRAW_REQUESTED_MSG` handler calls `clear_dirty` (disarming the vsync backstop), then parks the request in `redraw_requested`, which only `WM_PAINT` reads — and nothing invalidates (event_loop.rs:2697-2712). Needs the `RDW_INTERNALPAINT` re-arm the WM_PAINT arm has.
8. **Double-paint race** — vblank worker `swap(false)` + `RedrawWindow` can fire before the posted MSG dispatches (likely exactly when the queue is flooded — the PR's own premise), yielding two paints per request (vsync.rs:114-141).
9. **In-handler `request_redraw`** re-posts unpaced (clear-at-entry) and double-dispatches via the WM_PAINT re-arm — animation loops get 2× events with no vblank pacing.

## Also to fold in

- **Fix `WM_IME_SETCONTEXT` masking** — it masks `ISC_SHOWUICOMPOSITIONWINDOW` off **wParam**, but the flag lives in **lParam** (upstream winit fixed this; the fork predates it). The mask is currently a no-op, which means the OS composition window suppression on main has never worked — and this fix must ship *with* this PR, never before it, or Windows users lose their only visible composition UI.
- **Restore `DefWindowProc` for `WM_IME_STARTCOMPOSITION`/`ENDCOMPOSITION`** — verified against both wezterm (no STARTCOMPOSITION handler at all) and upstream winit (DefWindowProc for both): consuming them has no precedent and risks candidate-window placement under legacy IMM32 IMEs. With the SETCONTEXT mask fixed, suppression works the documented way.
- Check `PostMessageW` result (failure currently wedges the coalescing latch — one line), gate the MSG dispatch on window visibility, skip empty `Commit("")` on cancel-with-GCS_RESULTSTR-size-0 IMEs.
- Smaller: OSC 12 splits the block into two cursor colors; `PreeditOverlay::get` lacks a column bound; ImeCaret atlas key omits `cell_h`; overlay allocates rows×cols every frame (~190 KB at 300×80) — a sparse representation is O(preedit).

## What held up under attack

Run-cache keying (no poisoning either direction — verified against the main emit loop's hasher scheme), `Preedit::new` char-boundary hardening (fixes a real panic vector from macOS surrogate-split offsets), `BOOL_IS_CURSOR_GLYPH` semantics in both render backends, wide-char lead/spacer math, run-breaking and decoration suppression, damage on preedit change, the Hangul IME fixes surviving the merge, WM_PAINT validation, vsync thread lifecycle, and the ENDCOMPOSITION stale-preedit guard.

Original verification: `cargo check`/`clippy` clean, rio-window type-checks for `x86_64-pc-windows-msvc`, 182 rioterm tests pass.

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1940,6 +1940,22 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                         if route.window.screen.context_manager.current().ime.preedit()
                             != preedit.as_ref()
                         {
+                            // Force a full terminal repaint so the
+                            // preedit overlay's cells re-emit on the
+                            // next frame. Without this the grid keeps
+                            // its previous CPU buffers and the overlay
+                            // wouldn't refresh until something else
+                            // damages the row.
+                            route
+                                .window
+                                .screen
+                                .context_manager
+                                .current_mut()
+                                .renderable_content
+                                .pending_update
+                                .set_terminal_damage(
+                                    rio_backend::event::TerminalDamage::Full,
+                                );
                             route
                                 .window
                                 .screen
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -460,6 +460,11 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                             // Just mark dirty — damage will be extracted from
                             // the terminal when the renderer locks it.
                             ctx_item.val.renderable_content.pending_update.set_dirty();
+                            // Skip the scheduler timer (only firable from
+                            // `about_to_wait`, which the IME key-repeat
+                            // message flood starves) and request the
+                            // paint directly. Vblank pacing stays with
+                            // the VSync worker.
                             route.request_redraw();
                         }
                     }
--- a/frontends/rioterm/src/context/mod.rs
+++ b/frontends/rioterm/src/context/mod.rs
@@ -116,7 +116,6 @@ impl<T: EventListener> Context<T> {
             state: self.renderable_content.cursor.state.new_from_self(),
             content: self.renderable_content.cursor.content_ref,
             content_ref: self.renderable_content.cursor.content_ref,
-            is_ime_enabled: false,
         }
     }
 }
--- a/frontends/rioterm/src/context/renderable.rs
+++ b/frontends/rioterm/src/context/renderable.rs
@@ -27,7 +27,6 @@ pub struct Cursor {
     pub state: CursorState,
     pub content: char,
     pub content_ref: char,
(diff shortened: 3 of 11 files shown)


raphamorim:
Closing in favor of a from-scratch implementation with a corrected design (single-row slide-left layout, grapheme-cluster shaping, contrast-safe caret, lifecycle handling); the review punch list in the body above documents why. The Windows redraw-pump work will be redone separately.

PR #1850, opened 2026-08-10
rioterm: shape attached combining marks with their base glyph

raphamorim:
Workstream 0c of the mode-2027 scoping — a standalone rendering bug fix that ships on its own.

## The bug

`build_row_fg` hashes a cell's attached `zerowidth` codepoints into the run-cache key (so `e`+U+0301 and `e`+U+0302 don't alias) but never put them in the **shaping buffer** — only the base char was pushed. Every combining sequence in the grid renders as its bare base character: decomposed `é` draws as `e`, Devanagari conjuncts lose their matras, etc.

## The fix

- Attached marks are appended to the shaping buffer after their base on both platforms, so the shaper composes them (`emit_preedit_cluster` in #1849 already does this for IME text; this brings grid text to parity).
- The glyph→cell mapping moves to explicit per-cell starts on **both** platforms. The swash path walked `char_indices()` assuming one char per cell — with marks in the buffer that cursor would misattribute every following glyph. macOS already used a cell-starts table; the two walks now share one implementation (`run_cell_starts`: UTF-16 units on macOS, byte offsets on swash — swash reports `cluster.source.start` as byte offsets).
- VS15/VS16 stay excluded, mirroring `hash_combining`: presentation is already resolved into `font_id`, and feeding a variation selector to a text font's shaper only yields a notdef.
- Kitty Unicode-placeholder cells are excluded from the mark push — their `zerowidth` encodes image-slice coordinates, not text.

## Verification

174 rioterm tests pass (two new: shaping-buffer layout incl. selector skipping, and the cell-starts attribution walk). Clippy/fmt clean. The swash side compiles only on non-macOS — the ubuntu CI leg is the verifier for that path (local cross-checks are blocked by native build scripts). Visual check: `echo -e 'éx'` should render `éx`, not `ex`.

### DIFF
--- a/frontends/rioterm/src/grid_emit.rs
+++ b/frontends/rioterm/src/grid_emit.rs
@@ -1195,11 +1195,13 @@ pub struct GridGlyphRasterizer {
     // cell mapping.
     #[cfg(target_os = "macos")]
     run_utf16_scratch: Vec<u16>,
-    /// On macOS, `run_cell_starts[i]` is the offset (in UTF-16 code
-    /// units) where cell `i` of the run begins inside
-    /// `run_utf16_scratch`. Length = cells in the run. Used to walk
-    /// shaped glyphs back to the cell they belong to.
-    #[cfg(target_os = "macos")]
+    /// `run_cell_starts[i]` is the offset where cell `i` of the run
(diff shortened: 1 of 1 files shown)


raphamorim:

PR #1851, opened 2026-08-10
rio-vt: repair wide-pair seams after range mutations

raphamorim:
Workstream 0d of the mode-2027 scoping — fixes a pre-existing class of grid corruption. **Adversarially double-reviewed, then ported to ghostty's seam-targeted design with measured perf parity.**

## The bug

`write_cell` has always repaired wide-pair boundaries when a single-cell overwrite splits a pair. The range mutations never did: **ECH**, **DCH**, **ICH**, **EL**, and (found in review) **ED**'s partial-row clears all strand wide-pair halves at their range boundaries — a `Wide` lead rendering with a live cell in its right half, orphaned spacers, stale `LeadingSpacer` padding. Corrupts rendering, selection snapping, and copy/serialize. Alacritty 0.26 does no repair in any of these handlers (verified against source); ghostty repairs all of them.

## The design (ghostty's)

`split_wide_seam(line, col, blank)` — the port of ghostty's `Screen.splitCellBoundary`: clears the pair a boundary before `col` would cut (both halves — the xterm/ghostty whole-char-erase convention), reverts stale row-end `LeadingSpacer` padding when `col == columns`, and severs the previous-row wrap link when the mutation reaches columns 0/1. Each op names its exact seams before mutating:

- ECH / EL: both range edges
- DCH: range start, source end, row end (so a shift can't drag padding into the interior)
- ICH: insert point, row end, plus the last-shifted-cell cleanup (a lead whose spacer won't survive the shift — ghostty `insertBlanks`)
- ED Above/Below: the cursor-side edges of the partial row

Cleared cells take the op's blank but keep **WRAPLINE** — the flag marks the logical line for reflow and lives on the row's last cell, which the op never targeted (review finding: destroying it silently turned soft wraps into hard breaks on resize).

## Performance (measured, release, 320-column worst case)

| variant | per EL op |
|---|---|
| no repair (main) | 42 ns |
| whole-row sweep (first version) | 299 ns |
| **seam-targeted (this PR)** | **61 ns** |

O(1) per seam; the print hot path is untouched.

## History of this PR

1. First version: whole-row invariant sweep.
2. Adversarial review found a WRAPLINE regression and the ED coverage gap (plus a damage-index nit and DCH's interior-`LeadingSpacer` drag); fixed.
3. Ported to ghostty's seam-targeted model for perf parity — all eight regression tests pass unchanged across all three versions, so the observable semantics are pinned.

## Verification

Eight regression tests (ECH both halves, DCH orphan + dragged padding, ICH pushed-off-edge, EL stranded spacer, stale `LeadingSpacer` reversion, WRAPLINE preservation, ED Above/Below). Full rio-vt suite 437 green; librio + rioterm unaffected; clippy/fmt clean.

### DIFF
--- a/rio-vt/src/crosswords/mod.rs
+++ b/rio-vt/src/crosswords/mod.rs
@@ -1444,76 +1444,81 @@ impl<U: EventListener> Crosswords<U> {
     /// of its per-cell loop; the write is then a single packed store, and the
     /// cursor cell is resolved once (not twice) in the common narrow case.
     #[inline]
-    /// Repair wide-pair seams on `line` after a range mutation.
+    /// Clear the wide pair that a cell boundary before `col` would
+    /// split, plus the cross-row links a mutation at that boundary
+    /// severs. The ghostty model (`Screen.splitCellBoundary`): each
+    /// range mutation names the exact seams it cuts, O(1) per seam,
+    /// instead of sweeping the row afterwards.
     ///
-    /// A blanket blank (ECH / EL) or a cell shift (DCH / ICH) can
-    /// split a wide pair at the edges of the affected range: a `Wide`
-    /// lead stranded without its `Spacer`, a `Spacer` stranded without
-    /// its lead, or a `LeadingSpacer` at the end of the previous row
-    /// whose wide char at column 0 was cleared. Mirrors `write_cell`'s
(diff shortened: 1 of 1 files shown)


raphamorim:

PR #1853, opened 2026-08-10
rio-unicode: one crate, one Unicode version — width and graphemes

raphamorim:
Workstream 0a of the mode-2027 scoping — the last of the four foundation pieces.

## The problem it closes

The tree had a Unicode version skew: widths came from `unicode-width-16` (the alacritty fork, frozen at Unicode 16), while `unicode-segmentation` (used by the preedit work in #1849) ships Unicode 17. Width and segmentation disagreeing about what Unicode says is exactly the class of bug ghostty's unified `src/unicode/` module exists to prevent — this is rio's version of that module.

## Width half — same semantics, current data

Provenance mattered here: unicode-width 0.1.13+/0.2.x quietly changed per-char semantics (NUL→None, soft hyphen→0, conjoining jamo→0, even a width-3), which is why the tree pinned the fork in the first place. I audited three generator generations empirically — a test walks **every codepoint** against `unicode-width-16`:

- 0.2.2 base: 4,149 diffs (semantic overhaul — rejected)
- 0.1.14 base: 4,149 diffs (same modernized code — rejected)
- **0.1.11 base (this PR): 196 diffs — every one a genuine Unicode 16→17 data change** (new scripts' combining marks → 0, Tangut/Khitan/CJK additions and new emoji → 2)

The audit test stays in `tests/` as a permanent record of the drift. `lib.rs` is vendored from the fork itself; the generator (upstream `unicode.py` v0.1.11 lineage, patched for Unicode 17 fetch paths) is checked in under `scripts/`.

## Grapheme half — built for the grid

`is_break(prev: GraphemeClass, next: GraphemeClass, &mut BreakState) -> bool` — a pairwise state machine carrying the three things single-pair rules can't see: emoji ZWJ sequences (GB11), regional-indicator parity (GB12/13), and Indic conjunct linkage (GB9c, via InCB-refined classes). This is the shape mode-2027 grid code needs — clustering codepoints cell by cell against the previous cell, ghostty-style, without materializing strings. `Graphemes`/`GraphemeIndices` iterators wrap it for string callers (the preedit in #1849 can drop `unicode-segmentation` once rebased).

**All 766 UCD `GraphemeBreakTest.txt` conformance vectors pass**, generated into a test by the checked-in `scripts/grapheme.py`. A lockstep test asserts the grapheme data and width tables always report the same Unicode version.

## Wiring

The workspace `unicode-width` alias now points at `rio-unicode` (path dep), so every consumer — rio-vt's 128 KiB runtime width table, sugarloaf, rioterm — moves to Unicode 17 in one step with no source changes. Full downstream suites green: 441 rio-vt, 172 rioterm, 46 librio, 133 sugarloaf. Clippy/fmt clean.

Follow-ups unblocked: #1849 switches its segmentation here; the mode-2027 core (workstream 2 of the scoping doc) now has its tables and state machine waiting.


raphamorim:

PR #1866, opened 2026-08-12
fix(macos): restore titlebar double-click with hidden tab strip

### LINKED ISSUE #1786: Double-click on title bar doesn't expand window to fill screen
**Description:**

Double-clicking on the title bar doesn't expand it to occupy the screen space (while keeping the menu). Also, instead of showing the regular cursor, it shows a vertical-text cursor.

**Steps to reproduce:**

- Open Rio terminal
- Double-click on the title bar
- Observe that the window does not expand to fill the screen

**Expected behavior:**
I believe double-clicking the title bar would maximize/expand the window (while keeping the menu bar visible), similar to standard macOS window behavior.

**Actual behavior:**
Nothing happens besides changing the cursor to a vertical-text pointer instead of staying as the default cursor.

**Version:**
I'm using Rio 0.4.12 (0.4.12.20260722193443)

Note: I wasn't able to attach a screenshot of this, because the editing cursor disappears when I try to capture the screen during this interaction.

jxdones:
Closes #1786.

## Summary
- Route clicks in the hidden single-tab chrome through the existing window drag and double-click handler.
- Use the default cursor over the reserved macOS chrome area.
- Preserve the existing behavior on Linux and Windows.

## Testing
- cargo fmt --all -- --check
- cargo test -p rioterm (183 passed)

### DIFF
--- a/frontends/rioterm/src/application.rs
+++ b/frontends/rioterm/src/application.rs
@@ -1622,17 +1622,17 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
                     route.request_redraw();
                 }
 
-                // Only force the default cursor while the island is
-                // visible — when it's hidden (hide_if_single + single
-                // tab on macOS) the band at the top has no tabs to
-                // hover, and the I-beam from the terminal grid below
-                // should stay during top-edge drags.
+                // The macOS full-size content view keeps this band as custom
+                // window chrome even when hide-if-single hides the island.
+                // Other platforms only reserve it while the island is drawn.
                 use crate::renderer::island::ISLAND_HEIGHT;
                 let scale_factor = route.window.screen.sugarloaf.scale_factor();
                 let island_height_px = (ISLAND_HEIGHT * scale_factor) as f64;
                 let num_tabs = route.window.screen.ctx().len();
                 let nav = &route.window.screen.renderer.navigation;
-                if nav.island_visible(num_tabs) && y <= island_height_px {
+                let custom_chrome_visible = nav.island_visible(num_tabs)
+                    || (cfg!(target_os = "macos") && nav.is_enabled());
+                if custom_chrome_visible && y <= island_height_px {
                     route.window.winit_window.set_cursor(CursorIcon::Default);
                     return;
                 }
--- a/frontends/rioterm/src/screen/mod.rs
+++ b/frontends/rioterm/src/screen/mod.rs
@@ -2837,9 +2837,11 @@ impl Screen<'_> {
 
         let mouse_x_unscaled = mouse_x as f32 / scale_factor;
 
-        // Island isn't painted (hide_if_single + single tab on macOS).
-        // Nothing to click on, so let the caller route the event to the
-        // grid for selection / double-click maximize at the OS title bar.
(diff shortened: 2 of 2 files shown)


frontends/rioterm/src/screen/mod.rs
```diff
@@ -2848,6 +2850,13 @@ impl Screen<'_> {
             if self.is_close_press_tail(mouse_x_unscaled) {
                 return true;
             }
+
+            #[cfg(target_os = "macos")]
+            if !is_right_click {
```
raphamorim, line 2855: Right-clicks here fall through to the grid, which starts below the band, so the click acts on the first terminal row while the pointer is over chrome. The visible-island path consumes right-clicks in its chrome regions (the x_in_tabs < 0 and past-last-tab arms return true without an action). Consuming them here too would keep the hidden band consistent with the visible one:

```rust
#[cfg(target_os = "macos")]
{
    if !is_right_click {
        self.on_chrome_press(window, chrome_press);
    }
    return true;
}
```

frontends/rioterm/src/application.rs
```diff
@@ -1622,17 +1622,17 @@ impl ApplicationHandler<EventPayload> for Application<'_> {
...
+                // The macOS full-size content view keeps this band as custom
+                // window chrome even when hide-if-single hides the island.
+                // Other platforms only reserve it while the island is drawn.
                 use crate::renderer::island::ISLAND_HEIGHT;
                 let scale_factor = route.window.screen.sugarloaf.scale_factor();
                 let island_height_px = (ISLAND_HEIGHT * scale_factor) as f64;
                 let num_tabs = route.window.screen.ctx().len();
                 let nav = &route.window.screen.renderer.navigation;
-                if nav.island_visible(num_tabs) && y <= island_height_px {
+                let custom_chrome_visible = nav.island_visible(num_tabs)
```
raphamorim, line 1633: This predicate now exists in three places that must agree: here, the new branch in screen/mod.rs, and padding_top_from_config, which is the drift island_visible's own doc comment warns about. A helper on Navigation would pin them together, something like:

```rust
/// Whether the top band is reserved as window chrome this frame,
/// painted island or not. Must agree with padding_top_from_config.
pub fn chrome_band_reserved(&self, num_tabs: usize) -> bool {
    if cfg!(target_os = "macos") {
        self.is_enabled()
    } else {
        self.island_visible(num_tabs)
    }
}
```

Then this line and the screen branch both call it, and the cfg in screen/mod.rs collapses into the same expression.

raphamorim (commented):
Reviewed against the padding and input routing paths. The core is right: with hide-if-single on macOS, padding_top_from_config reserves ISLAND_HEIGHT whenever Tab navigation is enabled (the render-from-top early return is non-macOS only), so the band really is dead chrome and routing its clicks through on_chrome_press restores drag and double-click maximize with the same state machine the visible island uses. The cursor gate also matches the padding predicate exactly, and NativeTab/Plain modes are untouched since is_enabled() is Tab-only.

Two small requests inline, and one note: the UCRT64 CI failure is not yours. Our new PR smoke lane built the whole workspace in debug, and the cdylib targets blow the PE export-ordinal limit on windows-gnu. Fixed on main; a re-run should come back green.

PR #1886, opened 2026-08-21
fix(rioterm): keep quick-select hints synced while scrolling

mirsella:
Quick-select hints were tied to the viewport where hint mode started. Scrolling could leave stale labels or exit hint mode when the new viewport had no matches.

This refreshes hints whenever the viewport moves and keeps hint mode active across empty viewports. Activating quick-select also preserves the current scroll position.


mirsella:
Note: the failing `cargo fmt` step is pre-existing on current `main`, not introduced by this PR.

PR #1887, opened 2026-08-22
Icon resource compilation fix on Windows

milkowski:
This change fixes resource compilation on Windows for `cargo install rioterm` case.
In contrast to regular in-tree build `cargo build` the cargo installation build is done in ephemeral directory that cannot be simply navigated by relative path. The `CARGO_MANIFEST_DIR` environment variable must be used to point to the proper location.

### DIFF
--- a/rio-window/build.rs
+++ b/rio-window/build.rs
@@ -47,6 +47,23 @@ fn main() {
     load_app_icon();
 }
 
+#[cfg(target_os = "windows")]
+fn load_app_icon() {
+    let mut res = winres::WindowsResource::new();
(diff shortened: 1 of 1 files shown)


raphamorim:
Thanks for tracking this down. Superseded by #1898, which keeps your commit and moves the icon into the rio-window package itself: the published crate cannot include files outside its root, so the CARGO_MANIFEST_DIR parent path still misses the icon for crates.io installs. #1898 also downgrades a missing icon to a build warning so registry builds never hard-fail. Closing in favor of that one.

PR #1892, opened 2026-08-24
fix(kitty): draw Unicode placeholders after a=t + a=p,U=1 (placement id 0, GPU upload)

### LINKED ISSUE #1891: Kitty Unicode placeholders never drawn after a=t + a=p,U=1 (placement id 0 not resolved; pixels never uploaded)
### Summary

Unicode placeholder cells (U+10EEEE, `a=p,U=1`) are never drawn for the sequence that real clients emit — `a=t` (transmit only), then `a=p,U=1,p=N`, then placeholder cells whose foreground color carries the image id and that carry **no** underline color. The placement is registered and the cells reserve space, but nothing appears. This is the sequence emitted by [snacks.image](https://github.com/folke/snacks.nvim) (Neovim) and yazi.

Two independent causes, both in the virtual-placement path:

1. **Placement id 0 is not resolved to "any virtual placement".** `flush_run` (`frontends/rioterm/src/renderer/mod.rs`, same in `librio/src/render_state.rs`) looks up `(image_id, placement_id)` and then `(image_id, 0)`. The spec says for the underline color: *"if it's omitted or zero, the terminal may choose any virtual placement of the given image"* — kitty (`graphics.c`, "Find the first virtual image placement") and ghostty (`graphics_unicode.zig`, `placeholderTarget`) do exactly that. With a placement registered as `p=11` and cells without an underline color, Rio finds nothing.

   This is the common case, not an edge case: Neovim's TUI emits the underline color (SGR 58) only for cells that have an underline attribute (`src/nvim/tui/tui.c`, `if (has_any_underline && ...)`), and snacks.image does not underline its placeholder cells while registering the placement with `p=N`.

2. **After `a=t`, the virtual `a=p` path never uploads the pixels to the GPU.** `store_graphic` intentionally defers the upload ("pixel data is sent to GPU when a=p placement arrives"). The direct path does it in `place_kitty_overlay`, and `kitty_transmit_and_display` (`a=T,U=1`) does it explicitly — with a comment saying that otherwise "the placeholder cells render as blank space". The plain `a=p,U=1` branch in `place_graphic` calls `place_virtual_graphic` only, so the placement exists but the texture never does. (`a=T,U=1` — what `kitten icat` emits — works, which hides the bug.)

### Environment

Rio 0.5.26 (also `main` at `8e7cb9e`, identical), Windows 11 Pro 26200, Neovim 0.12.5 + snacks.nvim.

### Reproduction (no Neovim needed)

```python
import base64, sys, time
ESC, PH = "\x1b", chr(0x10EEEE)
DIA = [chr(int(h, 16)) for h in "0305,030D,030E,0310,0312,033D,033E,033F,0346,034A,034B,034C".split(",")]
IMG = "/absolute/path/to/some.png"
w = sys.stdout.write
w(f"{ESC}_Ga=t,q=2,f=100,t=f,i=101;{base64.b64encode(IMG.encode()).decode()}{ESC}\\")
w(f"{ESC}_Ga=p,U=1,q=2,i=101,p=11,c=12,r=4{ESC}\\")
w("Expected: the image in 12x4 cells below this line\n")
for r in range(4):
    w(f"{ESC}[38;2;0;0;101m" + "".join(PH + DIA[r] + DIA[c] for c in range(12)) + f"{ESC}[0m\n")
sys.stdout.flush(); time.sleep(15)
```

Rio 0.5.26: nothing is drawn. kitty/ghostty: the image appears.

Variants that isolate the two causes (`RIO_LOG_LEVEL=debug` confirms `Virtual placement: image_id=101, placement_id=11` is registered in all of them):

| transmit | placement | cell colors | result |
|---|---|---|---|
| `a=T` | `p=11` | fg + underline color 11 | drawn |
| `a=T` | `p=11` | fg only | **nothing** (cause 1) |
| `a=T` | no `p` | fg only | drawn |
| `a=t` | `p=11` | fg only | **nothing** (causes 1 + 2) |
| `a=t` | `p=11` | fg only, with cause 1 fixed | **nothing** (cause 2) |

### Fix

PR follows: `resolve_virtual_placement` in `kitty_virtual.rs` (non-zero id must match exactly, id 0 falls back to the placement with the lowest id for that image so the choice is deterministic), used by both renderers; and `upload_for_virtual_placement` in `crosswords/mod.rs`, called before the first virtual placement of an image (re-placing does not resend pixels; retransmission with virtual placements refreshes them). With both, snacks.image renders images and LaTeX math inline in Neovim on Windows.

reminiscience:
## Summary

Fixes #1891 — Unicode placeholder cells (`a=p,U=1` + U+10EEEE) were never drawn for the sequence real clients emit (`a=t`, then `a=p,U=1,p=N`, then cells without an underline color), e.g. snacks.image in Neovim and yazi. Two independent causes, one commit each:

1. **Resolve placement id 0 to any virtual placement of the image** (`kitty_virtual::resolve_virtual_placement`, used by both the rioterm renderer and `librio`). Per the spec the underline color carries the placement id and "if it's omitted or zero, the terminal may choose any virtual placement of the given image"; kitty and ghostty do that, Rio only tried the exact key and `(image_id, 0)`. Neovim's TUI emits SGR 58 only for underlined cells, so snacks' placeholder cells always arrive with id 0 while the placement is registered as `p=N`. A non-zero id must still match exactly; id 0 picks the lowest placement id so the result does not depend on hash-map order.

2. **Upload pixels before the first virtual placement after `a=t`** (`upload_for_virtual_placement`). `store_graphic` defers the GPU upload to the placement; the direct `a=p` path and `a=T,U=1` do it, the plain `a=p,U=1` branch did not, so the placement existed but the texture never did. Re-placing the same image (clients do that on every relayout) does not resend; retransmitting an image with virtual placements now refreshes them like it does for direct ones.

## Test plan

- [x] `cargo test -p rio-vt` — 3 new tests for `resolve_virtual_placement`, 1 for `upload_for_virtual_placement`, existing virtual-placement tests unchanged (35 pass in the filtered run)
- [x] Windows build with `--features wgpu`: the script from the issue draws the image; the four variants from the issue table all draw
- [x] Neovim 0.12.5 + snacks.image with `SNACKS_KITTY=1`: images and `pdflatex`-rendered math render inline in markdown buffers (previously: space reserved, nothing drawn)
- [ ] macOS/Linux not tested (the changed code is platform-independent)

### DIFF
--- a/frontends/rioterm/src/renderer/mod.rs
+++ b/frontends/rioterm/src/renderer/mod.rs
@@ -1107,10 +1107,11 @@ impl Renderer {
             z_index: i32,
             clip: (f32, f32, f32, f32),
         ) {
-            let vp = rc
-                .kitty_virtual_placements
-                .get(&(run.image_id, run.placement_id))
-                .or_else(|| rc.kitty_virtual_placements.get(&(run.image_id, 0)));
+            let vp = rio_backend::ansi::kitty_virtual::resolve_virtual_placement(
+                &rc.kitty_virtual_placements,
+                run.image_id,
+                run.placement_id,
+            );
             let vp = match vp {
                 Some(v) => v,
                 None => return,
--- a/librio/src/render_state.rs
+++ b/librio/src/render_state.rs
@@ -4,7 +4,8 @@ use rio_vt::ansi::graphics::{
     VirtualPlacement,
 };
 use rio_vt::ansi::kitty_virtual::{
-    compute_run_geometry, IncompletePlacement, PlaceholderRun, PLACEHOLDER,
+    compute_run_geometry, resolve_virtual_placement, IncompletePlacement, PlaceholderRun,
+    PLACEHOLDER,
 };
 use rio_vt::ansi::CursorShape;
 use rio_vt::config::colors::term::TermColors;
@@ -355,10 +356,11 @@ impl RenderState {
                      run: PlaceholderRun,
                      line: usize,
                      start_col: usize| {
-            let placement = graphics
-                .kitty_virtual_placements
-                .get(&(run.image_id, run.placement_id))
-                .or_else(|| graphics.kitty_virtual_placements.get(&(run.image_id, 0)));
+            let placement = resolve_virtual_placement(
+                &graphics.kitty_virtual_placements,
+                run.image_id,
+                run.placement_id,
+            );
             let Some(placement) = placement else { return };
             let Some(image) = graphics.get_kitty_image(run.image_id) else {
                 return;
--- a/rio-vt/src/ansi/kitty_virtual.rs
+++ b/rio-vt/src/ansi/kitty_virtual.rs
@@ -1,6 +1,8 @@
 // Kitty graphics protocol virtual placement encoding/decoding
 
+use crate::ansi::graphics::VirtualPlacement;
 use crate::config::colors::{AnsiColor, ColorRgb};
+use rustc_hash::FxHashMap;
 
 /// The Kitty Unicode placeholder codepoint (U+10EEEE). Cells containing
 /// this codepoint are interpreted as image placeholders; their fg color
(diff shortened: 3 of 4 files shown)


raphamorim:
Fixed on https://github.com/raphamorim/rio/pull/1893 . took your commit along! thank you

PR #1894, opened 2026-08-25
fix(rio-vt): close a synchronized update whose end marker arrives split

fglogan:
## What this fixes

A synchronized update (DEC mode 2026) could fail to close when the end marker (ESU) arrived across two reads.

The parser handled a chunk that contained a begin marker (BSU) partway through by arming the sync timeout, but it still consumed the rest of the chunk itself. The bytes after the BSU never entered the sync buffer, so if the ESU began at the end of that chunk and finished in the next one, `advance_sync_csi` never saw it whole. The update stayed open until the 150 ms timeout expired *and* more output arrived. A program that paints one frame and then waits for input never sends more output, so the pane kept showing the previous frame.

This shape is common: a TUI that writes BSU, the frame, and ESU as three separate flushes hits it whenever the read boundary falls inside the ESU.

## The change

Chunks are now cut right after the first BSU. The head goes through the parser and the tail through the sync path, exactly as if the BSU had ended the chunk. A small carry of the last parsed bytes lets the code recognise a BSU whose own start arrived in an earlier chunk. Both the CSI (`\e[?2026h` / `\e[?2026l`) and DCS (`\eP=1s\e\\` / `\eP=2s\e\\`) forms are handled.

One file changed: `rio-vt/src/performer/handler.rs`.

## Tests

Four new tests in the same file:

- mid-chunk BSU followed by a split ESU, CSI form
- the same, DCS form
- a split BSU followed by a split ESU
- the whole sequence fed one byte at a time, which must match any other chunking

`cargo test -p rio-vt` on this branch: 532 passed, 0 failed.

## How to reproduce without the fix

Feed `advance` two chunks: `b"abc\e[?2026hXYZ\e[?20"` and then `b"26l"`. Before the change the grid still shows the old content after the second chunk; with it, `XYZ` is on screen as soon as the second chunk lands.

### DIFF
--- a/rio-vt/src/performer/handler.rs
+++ b/rio-vt/src/performer/handler.rs
@@ -516,6 +516,12 @@ struct SyncState {
 
     /// Bytes read during the synchronized update.
     buffer: Vec<u8>,
+
+    /// The last bytes handed to the normal parser, at most one byte short
+    /// of a BSU. Lets `find_bsu_end` recognise a BSU whose start arrived
+    /// in an earlier chunk, so the bytes after it still reach the sync
+    /// buffer.
+    carry: Vec<u8>,
 }
 
 #[derive(Debug, Default)]
@@ -543,6 +549,7 @@ impl Default for SyncState {
     fn default() -> Self {
         Self {
             buffer: Vec::with_capacity(SYNC_BUFFER_SIZE),
+            carry: Vec::with_capacity(SYNC_ESCAPE_LEN),
             timeout: Default::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


raphamorim:
Thanks for the report and the precise repro, the bug was exactly as described. Superseded by #1902, which fixes it at the parser level instead: the parser stops right after dispatching a BSU and reports how many bytes it consumed, so the tail is buffered without rescanning chunks (the scan here measured 18-26% slower on escape-dense input, and the exact-form match missed multi-param BSUs like `ESC[?2026;1004h`). Same design upstream vte 0.15 shipped for this bug in alacritty; benchmarks on #1902 land within noise of main. Your test scenarios carry over, covered there plus an exhaustive chunking-equivalence sweep. Closing in favor of that one.

PR #1895, opened 2026-08-25
librio: expose bracketed paste via rio_surface_paste

ezemtsov:
**Problem.** librio's only C input entry point is `rio_surface_text`, which writes bytes raw. An embedder driving librio (e.g. an Emacs module) therefore can't perform a bracketed paste: it can neither send text atomically nor query mode 2004 to wrap the text itself — so a multi-line paste into a shell runs line by line.

**Solution.** `Surface::paste` already does this correctly (mode-2004 check, `\r` normalization, `ESC[200~…ESC[201~` markers), it just isn't reachable from C. Expose it as a thin `rio_surface_paste`, mirroring `rio_surface_text`. Purely additive; no behavior change.

### DIFF
--- a/librio/include/librio.h
+++ b/librio/include/librio.h
@@ -204,6 +204,8 @@ rio_surface_id_t rio_surface_id(const rio_surface_t *surface);
 
 /* Input entry points are callable from any thread. */
 void rio_surface_text(rio_surface_t *surface, const char *bytes, size_t len);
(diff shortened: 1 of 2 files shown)


raphamorim:
Thanks, this was the right call. Superseded by #1896, which keeps your commit as-is and adds escape filtering plus verbatim payload inside the brackets on top. Closing in favor of that one.

PR #1899, opened 2026-08-26
librio: shell-independent paste gate, fix C host SIGPIPE

raphamorim:
Follow-up to #1897, whose idle-prompt ordering passed locally but still failed on the macOS runner (found_paste=0). Two changes:

**Shell-independent paste check.** The gate now pastes into a second surface running /bin/cat instead of the login shell. There is no line editor in the way and the tty's own echo puts the text on screen, so the check no longer depends on which shell the runner uses or what its readline does with type-ahead.

**SIGPIPE fix in corcovado (the actual flake).** Stress-running the reworked gate surfaced a real embedding bug: roughly 1 in 8 runs died with exit 141. Backtrace: during Pty teardown, signal-hook's SignalDelivery closes its self-pipe socketpair; a SIGCHLD from the dying child lands mid-close and the signal handler writes to that pipe to wake readers. signal-hook ignores the write error on purpose, which assumes EPIPE comes back as an error return. That holds in Rust binaries (std ignores SIGPIPE at startup) but not in C or Swift hosts, where the default disposition kills the whole process. The fix sets SO_NOSIGPIPE at corcovado socket creation (Socket::new and Socket::pair, macOS/iOS/FreeBSD), the same thing upstream mio does on Apple targets, so every corcovado socket including the signal self-pipe returns EPIPE instead of raising.

With the fix the gate passes 40/40 stress runs where it previously failed ~5; the unmodified C gate doubles as a regression check that librio cannot SIGPIPE a C host on teardown. On Linux the corresponding hole (signal-hook's send without MSG_NOSIGNAL) has no per-fd opt-out and remains an upstream signal-hook assumption; non-Rust embedders there should ignore SIGPIPE, as long-lived C programs generally do.

### DIFF
--- a/corcovado/src/socket.rs
+++ b/corcovado/src/socket.rs
@@ -14,6 +14,28 @@ fn cvt(i: libc::c_int) -> io::Result<libc::c_int> {
     }
 }
 
+// Writers must get EPIPE back instead of SIGPIPE: these sockets are
+// written from signal handlers (teletypewriter's signal self-pipe),
+// and non-Rust hosts embedding librio do not ignore SIGPIPE, so the
+// default disposition would kill their process during fd teardown.
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "freebsd"))]
+unsafe fn set_nosigpipe(fd: c_int) -> io::Result<()> {
+    let on: c_int = 1;
+    cvt(libc::setsockopt(
+        fd,
+        libc::SOL_SOCKET,
+        libc::SO_NOSIGPIPE,
+        &on as *const c_int as *const libc::c_void,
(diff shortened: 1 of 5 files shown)


raphamorim:
Review follow-up in 44df1d3: the SO_NOSIGPIPE now lives in corcovado's Socket::new/Socket::pair (error-checked, next to the existing FIOCLEX/FIONBIO setup, matching what upstream mio does on Apple targets) instead of being patched one layer up in teletypewriter, which is reverted. This also covers every other corcovado socket, not just the signal self-pipe. The C gate lost its second fixed sleep (pty input queues in the line discipline regardless of when cat spawns), creates the render state before pasting, and no longer leaks on the paste-surface failure path. Stress: 40/40 passes.

PR #1900, opened 2026-08-26
librio tests: spawn sleeper via /bin/sh

raphamorim:
The Nix Build workflow has been failing on main for several days: 11 librio tests spawn /bin/sleep through the quiet_surface helper, and the nix build sandbox provides no /bin/sleep (only /bin/sh). The sleeper now spawns as /bin/sh -c 'sleep 300', which works in the sandbox and everywhere else.

Surfaced on #1898, whose Cargo.lock change happened to trigger the nix workflow on a PR; the failure predates it (every recent Nix Build run on main is red with the same 11 tests).

### DIFF
--- a/librio/src/lib.rs
+++ b/librio/src/lib.rs
@@ -1584,15 +1584,16 @@ mod tests {
     }
 
     // A child that never writes, so buffer contents are exactly what the
-    // test injected and comparisons can't race the shell prompt.
+    // test injected and comparisons can't race the shell prompt. Spawned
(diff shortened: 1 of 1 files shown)


raphamorim:
Validated pre-merge via workflow_dispatch on this branch (the nix workflow's PR path filter skips .rs-only changes): https://github.com/raphamorim/rio/actions/runs/32949461027 — both legs pass, first green Nix Build since the failures began.

PR #1905, opened 2026-08-28
fix: window.opacity transparency on Windows

raphamorim:
Supersedes #1583, keeping @shiena's two commits as-is and addressing the review findings on top.

shiena's diagnosis and fix are exactly right and match what the references converged on: DWM composites premultiplied alpha (winit's with_transparent uses the same DwmEnableBlurBehindWindow empty-region trick wezterm hand-rolls to enable per-pixel alpha), alacritty premultiplies its clear color by opacity the same way, kitty premultiplies pervasively including its clear color, and ghostty premultiplies every color in-shader with ONE/ONE_MINUS_SRC_ALPHA blending.

What this adds on top:

**Alpha-capable adapter preference (the remaining gap).** wgpu's DX12 backend exposes only Opaque composite alpha for plain HWND swapchains (PreMultiplied needs a DirectComposition target), and rio's Windows default is Backends::all(), so whether transparency worked depended on whether the adapter request happened to land on Vulkan (shiena's test) or DX12 (wezterm's long-open WebGPU-frontend transparency bug, wezterm#6265). When window.opacity < 1, sugarloaf now prefers the highest-class adapter whose surface offers PreMultiplied or PostMultiplied, falling back to the normal request when none exists. Backend-agnostic, no double surface creation, no panic on Vulkan-less machines.

**CPU blend ALU cost removed.** The fourth channel had added a third multiply per blend. A now rides the lane pair G occupied alone (00AA00GG mirroring 00RR00BB), so all four kernels (scalar SWAR, x4/x8 const-src, var-src) are back to two multiplies per pixel, the same count as the old opaque three-channel code. The alpha channel is literally free.

**Tests locking the invariant.** Seven new unit tests: exhaustive opaque-destination-stays-opaque over all 256 source alphas, transparent-source identity, Porter-Duff alpha composition, exact bit-for-bit agreement between the scalar SWAR and all three SIMD kernels over 2000 randomized valid-premul pairs, output-stays-valid-premul, plus the same alpha assertions for the scalar blends in grid/cpu.rs and text.rs.

**cfg cleanup plus one missed spot.** The triplicated any(feature = "wgpu", target_os = "windows") is now a single wgpu_backend alias in rioterm's build.rs (cfg_aliases, same pattern as rio-window). Converting every site uniformly also surfaced one the original PR missed: update_filters on config reload was still gated on the bare feature, so default Windows builds ignored [renderer] filters edits until restart.

One dependency worth recording: the CPU path's alpha relies on softbuffer copying the top byte through to the DWM redirection surface, which its 0.4 docs leave unspecified (they say the top 8 bits are to be set to 0). It works, shiena verified on Windows 11, and softbuffer master is growing an AlphaMode API worth adopting once premultiplied gains Windows support.

Validation: 140 sugarloaf tests pass, clippy clean, cargo check for x86_64-pc-windows-msvc with default features and with --features wgpu both pass, fmt clean. The blend-kernel changes are covered by the exact-agreement tests; the adapter preference needs a Windows machine with both DX12 and Vulkan to observe, which CI cannot do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4042,6 +4042,7 @@ version = "0.5.26"
 dependencies = [
  "ahash",
  "bitflags 2.13.1",
+ "cfg_aliases",
  "clap",
  "corcovado",
  "cpal",
--- a/frontends/rioterm/Cargo.toml
+++ b/frontends/rioterm/Cargo.toml
@@ -60,6 +60,9 @@ cpal = { version = "0.15", optional = true }
 [target.'cfg(target_os = "macos")'.dependencies]
 objc = { workspace = true }
 
+[build-dependencies]
+cfg_aliases = "0.2.1"
+
 [target.'cfg(windows)'.build-dependencies]
 winres = "0.1.12"
 
--- a/frontends/rioterm/build.rs
+++ b/frontends/rioterm/build.rs
@@ -1,9 +1,16 @@
 fn main() {
     cfg_aliases::cfg_aliases! {
-        // rioterm's own `wgpu` feature, or Windows, where sugarloaf and
-        // rio-backend get the feature through the target-specific
-        // dependency override so the wgpu code paths always exist.
-        wgpu_backend: { any(feature = "wgpu", target_os = "windows") },
+        // rioterm's own `wgpu` feature, or the targets whose
+        // target-specific dependency overrides force the feature onto
+        // sugarloaf and rio-backend (Windows and wasm32 in Cargo.toml),
+        // so the wgpu code paths always exist there.
+        wgpu_backend: {
+            any(
+                feature = "wgpu",
+                target_os = "windows",
+                target_arch = "wasm32"
+            )
+        },
     }
 
     // wgpu-hal references HDR colorspace constants (ExtendedDisplayP3,
--- a/frontends/rioterm/build.rs
+++ b/frontends/rioterm/build.rs
@@ -1,4 +1,11 @@
 fn main() {
+    cfg_aliases::cfg_aliases! {
+        // rioterm's own `wgpu` feature, or Windows, where sugarloaf and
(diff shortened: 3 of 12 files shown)


raphamorim:
Independent double-review found ten issues; all fixed in a859ca7 (plus the glass predicate in the same commit). The two most important were behavioral regressions on non-Windows paths that this PR would have introduced:

- The wgpu clear was premultiplied for every alpha mode except PostMultiplied, darkening opaque surfaces (Auto resolving to Opaque, e.g. Vulkan-less Windows falling back to DX12) with no transparency in return. Now premultiplied only when the resolved mode is PreMultiplied.
- The CPU background premultiply reached macOS and Wayland, where softbuffer drops the alpha byte (CGImage NoneSkipFirst, Xrgb8888), turning opacity into pure dimming and glass mode into solid black. Now gated to Windows, the only softbuffer presenter that forwards alpha to the compositor.

Also fixed: alpha-mode selection now prefers PreMultiplied over PostMultiplied (every sugarloaf pipeline emits premultiplied output, so PostMultiplied surfaces double-attenuated translucent pixels); the preferred-adapter path falls back to the default request instead of aborting when a broken ICD refuses device creation; software rasterizers can no longer win adapter selection and ties resolve deterministically; adapters are enumerated once instead of twice per startup; a one-shot warning fires when [renderer] filters (whose final pass overwrites frame alpha) silently cancel window.opacity; the wgpu_backend alias covers wasm32, whose dependency override also forces the feature; the transparency predicate reuses window_should_be_opaque so glass blur engages it; and the triplicated pixel/blend helpers are consolidated into a shared crate::premul module, removing the stale 0x00RRGGBB comment that contradicted the code beneath it.

Sugarloaf tests, clippy, fmt, and cargo check for host, x86_64-pc-windows-msvc, --features wgpu, and libsugarloaf all pass.

PR #1922, opened 2026-09-06
Surface terminal bells on tabs, with a font-free indicator

raphamorim:
Supersedes #1881. Carries @aymanbagabas's three commits verbatim (bell tab indicator, immediate tab titles, BEL parser test), with review fixes layered on top.

A `BEL` from a background tab now flags its tab in the strip, and OSC 0/2 title changes reach the tab strip immediately as events. See #1881 for the full design write-up. Gated behind `bell.tab-indicator` (default `true`).

Titles are now purely event-driven: the standing 2s poll is gone and title rendering never inspects the foreground process. `{{ title }}` comes from OSC 0/2 events, the path variables come from OSC 7 events (empty for shells without that integration), `{{ program }}` is the name of the command the pane spawned, and `{{ columns }}`/`{{ lines }}` re-render on resize. An idle terminal does zero title work and no timer ever wakes the process.

### DIFF
--- a/rio-backend/src/config/bell.rs
+++ b/rio-backend/src/config/bell.rs
@@ -1,15 +1,20 @@
+use crate::config::defaults::default_bool_true;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Bell {
     #[serde(default = "default_audio_bell")]
     pub audio: bool,
+    /// Mark background tabs that rang the bell with a 🔔 in the tab strip.
+    #[serde(default = "default_bool_true", rename = "tab-indicator")]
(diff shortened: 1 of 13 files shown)


rio-backend/src/config/bell.rs
```diff
@@ -1,15 +1,20 @@
+use crate::config::defaults::default_bool_true;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Bell {
     #[serde(default = "default_audio_bell")]
     pub audio: bool,
+    /// Mark background tabs that rang the bell with a 🔔 in the tab strip.
```
aymanbagabas: Stale comment?

PR #1927, opened 2026-09-08
fix: preserve Unix PTY lifecycle and final output

### LINKED ISSUE #1926: Unix PTY Lifecycle Can Signal Reused PIDs and Drop Final Output
## Summary

At rio revision `7ae087500bcde5c0c9f09cb9c50382e9220b3360`, the Unix PTY path has two independently actionable lifecycle defects:

1. a reaped child can later be signalled through its saved numeric PID;
2. final PTY output can be lost during exit under terminal-state lock contention.

## Final-Output Failure

A real PTY child can write a final marker, exit 0, emit child-exit/close, and leave the marker absent from the final frame. This is reproduced with coordinated snapshot-lock contention (timing-dependent).

`Machine::pty_read` buffers bytes locally. After failed lock acquisition it may read again; Linux PTY hangup may then yield `EIO`. The error path returns before the already-read bytes reach the parser. This mechanism is source-backed and consistent with the failure; the regression does not prove the exact syscall sequence.

## Child Lifecycle Risk

`Child::waitpid(&self)` reaps without recording completion; `Child::drop` unconditionally sends `SIGHUP` to the saved PID. PID reuse can therefore signal an unrelated process. This is a source-level risk, not an observed incident. PID reuse does **not** cause the output-loss failure.

## API Boundary

`Pty.child` and `Child.pid` are public, but `Child.process`, `Machine.pty`, and `State.parser` are private. There is no small ownership-safe repair while retaining `Machine<Pty, Listener>`. A custom EIO-to-EOF reader only mitigates one path.

## Expected Repair

- Make termination/reap explicit, idempotent, status-preserving, and incapable of signalling after reap.
- Parse/preserve all buffered bytes before EOF/HUP/error; publish exit only after final drain and pending synchronized updates.
- Define bounded handling for descendants holding the slave open.

## Regression Coverage

Add a synthetic `EventedPty` bytes → EIO test under lock contention; EOF/HUP residual output; parsing-budget and pending-sync cases; plus real-child natural exit, repeated close, ignored SIGHUP, reap completion, setup/reader failures, and no post-reap signal.

Happy to provide the embedding-side regression harness.

rawkode:
## Summary

Fixes #1926.

Unix PTY children now retain their exit status and serialize termination with reaping, preventing subsequent shutdown or Drop from signaling a reaped PID. Shutdown sends SIGHUP, allows a 100 ms grace period, escalates to SIGKILL when necessary, and reaps the child. Frontend destructors delegate shutdown to the PTY owner instead of signaling saved PIDs directly. Setup failures also release descriptors and terminate owned children. Child ownership lives in a dedicated module with explicit running/exited states; a retired child retains no signalable PID. External reaping consistently reports an unavailable status rather than appearing to be running.

The reader parses buffered bytes before returning EOF or errors, drains residual output on HUP, and continues across parsing budgets after child exit. One finalization path handles setup failures, I/O failures, explicit shutdown, and natural exit. Pending synchronized updates are flushed before exit notification. Final draining stops when a read would block and limits continuous descendant output to 100 ms between batches; lock waits, parsing, and final reap can extend total completion time.

## Validation

Using Rust 1.98.0 on macOS:

- `cargo +1.98.0 test -p rio-vt -p teletypewriter --lib --quiet`: 541 terminal-core tests and 15 PTY tests passed.
- `cargo +1.98.0 test -p librio --no-default-features --features pty --lib --quiet`: 53 embedding tests passed, including real PTY shells.
- `cargo +1.98.0 check -p rio-vt --no-default-features --quiet`, formatting, and `git diff --check` passed.
- Regression coverage includes deterministic bytes-to-EIO under terminal-lock contention, EOF/HUP residual output, multiple parsing budgets, synchronized-update notification ordering, setup/reader failure cleanup, cached status, ignored SIGHUP, and a sentinel process verifying no signal after reap.
- Restoring the original early error return makes the EIO regression fail with the final marker absent; it passes with the fix restored.

Linux native CI passed on the preceding commit `0c53582ed0`, including the HUP-only drain regression and a new epoll injected-event roundtrip test. That test exposed a conversion path that dropped HUP/error readiness; injected epoll events now preserve both flags. EIO under coordinated lock contention remains synthetic coverage. Independent read-only review found no blocking issues.

The `EventedPty::shutdown` hook is explicitly optional: Unix implements termination/reaping, while the default is a no-op and Windows retains cleanup on drop. Regression tests also verify exactly-once finalization and external-reap event delivery.

### DIFF
--- a/corcovado/src/sys/unix/epoll.rs
+++ b/corcovado/src/sys/unix/epoll.rs
@@ -268,8 +268,19 @@ impl Events {
     }
 
     pub fn push_event(&mut self, event: Event) {
+        // Unlike registration interests, injected events must retain HUP and
+        // error readiness. The kernel reports these flags automatically for
+        // file descriptors, but user-space registrations have no kernel event.
+        let readiness = UnixReady::from(event.readiness());
+        let mut events = ioevent_to_epoll(event.readiness(), PollOpt::empty());
+        if readiness.is_hup() {
+            events |= EPOLLHUP as u32;
+        }
+        if readiness.is_error() {
+            events |= EPOLLERR as u32;
+        }
         self.events.push(libc::epoll_event {
-            events: ioevent_to_epoll(event.readiness(), PollOpt::empty()),
+            events,
             u64: usize::from(event.token()) as u64,
         });
     }
@@ -296,4 +307,24 @@ pub fn millis(duration: Duration) -> u64 {
         .as_secs()
         .saturating_mul(MILLIS_PER_SEC)
         .saturating_add(millis as u64)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn injected_events_preserve_hangup_and_error_readiness() {
+        for readiness in [
+            Ready::from(UnixReady::hup()),
+            Ready::from(UnixReady::error()),
+            Ready::readable() | UnixReady::hup() | UnixReady::error(),
+        ] {
+            let mut events = Events::with_capacity(1);
+            events.push_event(Event::new(readiness, Token(42)));
+            let event = events.get(0).unwrap();
+            assert_eq!(event.readiness(), readiness);
+            assert_eq!(event.token(), Token(42));
+        }
+    }
 }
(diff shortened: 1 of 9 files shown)


raphamorim:
Took the changes, and did few updates here https://github.com/raphamorim/rio/pull/1930 . merged in main! thank you

